PR #19, opened 2016-12-31
improve openssl speed

ghost:
Hi,

I've found a way to make openssl key generation faster. 
According to this post on Stackexchange
[linux - Diffie Hellman parameters still calculating after 24 hours - Information Security Stack Exchange](http://security.stackexchange.com/questions/95178/diffie-hellman-parameters-still-calculating-after-24-hours/95184#95184)

it makes no difference regarding security.


angristan:
If you want faster DH key generation, please use 2048 bits keys.

PR #26, opened 2017-01-14
Missing "-get" for OpenVPN install on Debian 8.x

robiiinos:
Simple mistake which is not currently affecting performance or updating on Debian 8.x

The 3 others custom installations for lastest OpenVPN release got the "apt-get"; so it's only visual


angristan:
Excuse me but... What ?

yolateng0:
From man apt:

    "APT is a management system for software packages."
no probleme to use apt or apt-get
From man apt-get:

    "apt-get is the command-line tool for handling packages."

So apt-get is a commandline interface for using apt.

angristan:
You should use apt now, no apt-get anymore and there are no differeneces between the two commands...

PR #33, opened 2017-02-06
OpenVPN 2.4 update

angristan:
As OpenVPN 2.4 is out, some improvements will be implemented to the script.

- [x] Use AES-GCM for the data channel's cipher instead of AES-CBC
- [x] Disable DH
- [x] Enable ECDH
- [x] Use an ECDSA certificate instead of RSA
- [x] [Use --tls-crypt instead of --tls-auth](https://community.openvpn.net/openvpn/changeset/c6e24fa3e16c32f9b427e360fd07102f613aa5c6/)
- [x] ~~[Update the Arch Linux installation ](https://www.archlinux.org/news/openvpn-240-update-requires-administrative-interaction/)~~ Remove Arch Linux support
- [x] Not related to the 2.4 update, but I want to use the best encryption settings everywhere (AES 256 GCM and SHA 512), thus I'll remove the choice for other ciphers, thus making the script more simple to use.
- [ ] Update the Readme

Also, we have to make sure that OpenVPN 2.4 is available everywhere.
- [x] Debian 7 (thanks to the [official OpenVPN repo](https://community.openvpn.net/openvpn/wiki/OpenvpnSoftwareRepos))
- [x] Debian 8 (thanks to the [official OpenVPN repo](https://community.openvpn.net/openvpn/wiki/OpenvpnSoftwareRepos))
- [x] Ubuntu 12.04 LTS (thanks to the [official OpenVPN repo](https://community.openvpn.net/openvpn/wiki/OpenvpnSoftwareRepos))
- [x] Ubuntu 14.04 LTS (thanks to the [official OpenVPN repo](https://community.openvpn.net/openvpn/wiki/OpenvpnSoftwareRepos))
- [x] Ubuntu 16.04 LTS (thanks to the [official OpenVPN repo](https://community.openvpn.net/openvpn/wiki/OpenvpnSoftwareRepos))
- [ ] Ubuntu 16.10
- [ ] CentOS 6
- [ ] CentOS 7
- [x] Arch Linux

Clients : 
- [x] Windows
- [x] MacOS (via [Homebrew](https://github.com/Homebrew/homebrew-core/blob/master/Formula/openvpn.rb) or [TunnelBlick](https://tunnelblick.net/))
- [ ] iOS
- [x] Android with [an alternative client](https://play.google.com/store/apps/details?id=it.colucciweb.free.openvpn)
-  For Linux clients, see above.

PR #39, opened 2017-02-23
Slight crypto upgrade.

NSAKEY:
SHA384 and SHA512 take roughly the same amount of time to run as SHA256. 

AES-256 vs AES-128 may actually impact performance, but I'm always in favor of picking the most hardcore options.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -466,7 +466,7 @@ auth SHA512
 $CIPHER
 tls-server
 tls-version-min 1.2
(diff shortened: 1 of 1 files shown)


angristan:
I will run more tests with the upcoming OpenVPN 2.4 update, and see how it impacts performance. If it does not affect speed too much, I'll choose the most "hard" parameters. Stay tuned with PR https://github.com/Angristan/OpenVPN-install/pull/33. :)

PR #45, opened 2017-03-23
Initial IPv6 support inside the tunnel for debian and ubuntu

Qwox:
I added some basic IPv6 support for Debian and Ubuntu systems. It works for additional routed subnets, parts of the primary subnet (except debian 7) and NAT for ip6tables (with suitable kernel and ip6tables).

I tested it on different vultr vm's and it runs pretty well. It does not really work on OpenVZ except for a routed subnet maybe, but i dont have such a machine, so i can't test it.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -151,6 +151,10 @@ if [[ -e /etc/openvpn/server.conf ]]; then
 					sed -i "/iptables -I FORWARD -m state --state RELATED,ESTABLISHED -j ACCEPT/d" $RCLOCAL
 				fi
 				sed -i '/iptables -t nat -A POSTROUTING -s 10.8.0.0\/24 -j SNAT --to /d' $RCLOCAL
+				# Remove ip6tables
+				sed -i "/ip6tables -I FORWARD -i tun+ -j ACCEPT/d" $RCLOCAL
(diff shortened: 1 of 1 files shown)


KeiroD:
Tested the IPv6 addition -- seemed to work ok, but unfortunately I have yet to get it to work successfully, as on the client-side, I'm getting the error:

```
root@EmberLaptop:~# openvpn --config emberlaptop.ovpn 
Sun Nov 12 21:58:23 2017 Unrecognized option or missing parameter(s) in emberlaptop.ovpn:15: block-outside-dns (2.3.10)
Options error: Unrecognized option or missing parameter(s) in emberlaptop.ovpn:20: tls-crypt (2.3.10)
Use --help for more information.
```

angristan:
This is because it uses the 2.4 branch as a base. Use a 2.4 client to support tls-crypt.

KeiroD:
Yep. I'll report back when I've got my vpn client updated to 2.4.

ghost:
Yep. I'll report back when I've got my vpn client updated to 2.4.

ghost:
@Angristan Is this deprecated and the #238 the updated PR? If yes it can be closed, too.

angristan:
They have not the same goal since this is for real IPv6 support and not NAT. But anyway it's too old and would be too hard to manage for now, so I'm closing it

Issue #25, opened 2017-01-12
Revoking a existing using cert making other cert disable too


robiiinos:
Hi,

I've found something by using your script. After adding one or many users with your scripts; when I try to revoke a existing user; it revoke him but also make all other cert unaivalable to use. Any other cert remaining on the current server are like "disabled" and even deleting all cert and re-making them one by one doesn't func...

[French]
Bonjour;
J'ai vu en utilisant ton script et après avoir ajouter plusieurs utilisateurs; que lorsqu'on en revoke un seul, tous les autres cert sont unitilsables... Plus aucun ne marche peut importe la configuration :(
Même en supprimant toutes les cert et en les re-créant; cela ne marche pas...

PS : J'attend impatiemment la màj pour OpenVPN 2.4.0 pour Debian et les DNS locaux 👍

robiiinos:
Solved by myself adding : chown nobody:nogroup /etc/openvpn/crl.pem
In the step when revoking a user cert; like idk why it wasn't func on my server if other do 👎

angristan:
Was it on your side or is it a script's bug ?

angristan:
Apparently it is a script's bug https://github.com/Angristan/OpenVPN-install/issues/32

borowskiio:
Hi.

It turns out that when revoking a client, script manipulates a crl.pem file - and it gives it wrong permissions of being readable only by the owner.

I added read permission for the group and others (as per dh.pem) and it went through.

I noticed it after reading through a syslog and a line:

`ovpn-server[3668]: xxx.xx.xx.xxxx:3987 CRL: cannot read: crl.pem`

Perhaps the script needs setting the umask at some point?

PR #47, opened 2017-03-28
chown crl.pem to nobody on revoke

tes5884:
This is a patch for [Issue #25](https://github.com/Angristan/OpenVPN-install/issues/25).

Apologies if I didn't format things correctly, this is my first time doing a pull request.

Thanks

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -133,6 +133,7 @@ if [[ -e /etc/openvpn/server.conf ]]; then
 			rm -rf pki/issued/$CLIENT.crt
 			rm -rf /etc/openvpn/crl.pem
 			cp /etc/openvpn/easy-rsa/pki/crl.pem /etc/openvpn/crl.pem
+			chmod nobody:nobody /etc/openvpn/crl.pem
 			echo ""
 			echo "Certificate for client $CLIENT revoked"
 			echo "Exiting..."
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -81,6 +81,13 @@ if [[ "$IP" = "" ]]; then
 	IP=$(wget -qO- ipv4.icanhazip.com)
 fi
 
+# Find out if the machine uses nogroup or nobody for the permissionless group
+if grep -qs "^nogroup:" /etc/group; then
+	NOGROUP=nogroup
+else
+	NOGROUP=nobody
+fi
+
 if [[ -e /etc/openvpn/server.conf ]]; then
 	while :
 	do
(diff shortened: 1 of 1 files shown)


openvpn-install.sh
```diff
@@ -133,6 +133,7 @@ if [[ -e /etc/openvpn/server.conf ]]; then
 			rm -rf pki/issued/$CLIENT.crt
 			rm -rf /etc/openvpn/crl.pem
 			cp /etc/openvpn/easy-rsa/pki/crl.pem /etc/openvpn/crl.pem
+			chmod nobody:nobody /etc/openvpn/crl.pem
```
Kcchouette: It would be better to use instead `chmod nobody:$NOGROUP /etc/openvpn/crl.pem` but 
```
# Find out if the machine uses nogroup or nobody for the permissionless group
if grep -qs "^nogroup:" /etc/group; then
        NOGROUP=nogroup
else
       	NOGROUP=nobody
fi
```
from [here](https://github.com/Angristan/OpenVPN-install/blob/89925cbbe854741d75f370be0525e8cf64b6c961/openvpn-install.sh#L383) should be moved on the top of the script

Kcchouette: Moreover, I think it's `chown`, not `chmod`, doesn't it?

tes5884: 1) Much better!
2) I'm an idiot - the irony is on my local (not git) script I actually did use chown...

Excuse my newbness - do I change my patch, or do you just submit your own? how does this work?

Thanks!

Kcchouette: You should change your patch or this owner of this repo can.
I'm not the owner so I can't push code on this repo.

So the best way is you change your patch :)

Thank you

tes5884:
Did the change - hopefully I did it correctly  :)

Thank you @Kcchouette for your assistance, and advice!

robiiinos:
I'm gonna try it!
Before your PR, I did a pretty same correction on my side to make it works.

Kcchouette:
Have you tried it, @Bashilor ?

chocolateshirt:
@Kcchouette : I had done it and it proved

go to /etc/openvpn

then make **chown nobody crl.pem**

angristan:
I tested using @tes5884 fork, but after the installation, crl.pem is still owned by root:root. I'm searching what is wrong

angristan:
Oh, ok.
Why do you put `chown nobody:$NOGROUP /etc/openvpn/crl.pem` after revoking the client ? Shouldn't it be when installing OpenVPN ?

Kcchouette (commented):
Indeed, you should replace that by `chown nobody:$NOGROUP /etc/openvpn/crl.pem`, shouldn't you?

angristan:
What ?

Kcchouette:
Replace `chmod 644 /etc/openvpn/crl.pem` by `chown nobody:$NOGROUP /etc/openvpn/crl.pem` (or adding it after) [here](https://github.com/tes5884/OpenVPN-install/blob/76548b00893a90b4823594002ac9f9ffce7d17e1/openvpn-install.sh#L417)?

PR #75, opened 2017-06-25
adding support for debian 9 stretch

goodtiding5:
There are two issues with debian 9 stretch:

1. VERSION_ID = 9, which is addressed in this patch; and
2. /etc/rc.local is no present.  We may have to export the firewall rules to a file, or considering to add a startup script for openvpn to run.


angristan:
The script is not compatible with Debian 9 yet.

angristan:
Concerning your second concern, it has been fixed : https://github.com/Angristan/OpenVPN-install/commit/ac203dd5eef586fb4a1c5db2fb0ca602d91edd32

goodtiding5:
Here is the thing: unless there is a rc-local.serivce under systemd, /etc/rc.local won't get invoked.  Please see this reference: https://www.linuxbabe.com/linux-server/how-to-enable-etcrc-local-with-systemd

Issue #60, opened 2017-05-06
OpenVPN removal keeps iptables rules


claymore008:
After removing (option 3), all placed firewall rules / iptables rules remains. 
Remove sequence doesnt delte created iptables rules/chains. Please implement ...

PR #83, opened 2017-07-17
Remove the use rc.local and insall iptables systemd service for all

DrXala:
Pull request for:

 * Install iptables systemd service for Debian, Ubuntu and CentOS
 * Fix iptables install for ArcLinux
 * Remove the use rc.local file
 * Remove all iptables rules when removing openvpn (cf. #60 )

I tested on Debian9 freshly installed and it's work fine. I dont have such a machine, so i can't test it. Maybe later...

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -24,7 +24,6 @@ if [[ -e /etc/debian_version ]]; then
 	# Getting the version number, to verify that a recent version of OpenVPN is available
 	VERSION_ID=$(cat /etc/os-release | grep "VERSION_ID")
   IPTABLES='/etc/iptables/iptables.rules'
-	RCLOCAL='/etc/rc.local'
 	SYSCTL='/etc/sysctl.conf'
 	if [[ "$VERSION_ID" != 'VERSION_ID="7"' ]] && [[ "$VERSION_ID" != 'VERSION_ID="8"' ]] && [[ "$VERSION_ID" != 'VERSION_ID="9"' ]] && [[ "$VERSION_ID" != 'VERSION_ID="12.04"' ]] && [[ "$VERSION_ID" != 'VERSION_ID="14.04"' ]] && [[ "$VERSION_ID" != 'VERSION_ID="16.04"' ]] && [[ "$VERSION_ID" != 'VERSION_ID="16.10"' ]] && [[ "$VERSION_ID" != 'VERSION_ID="17.04"' ]]; then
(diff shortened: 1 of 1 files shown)


DrXala:
Ref for change iptables.service:
https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/
https://www.freedesktop.org/software/systemd/man/systemd.special.html
https://bugs.archlinux.org/task/33478
http://unix.stackexchange.com/a/353852/163362

angristan:
Is 

```
iptables -F
iptables -X
iptables -t nat -F
iptables -t nat -X
iptables -t mangle -F
iptables -t mangle -X
iptables -P INPUT ACCEPT
iptables -P FORWARD ACCEPT
iptables -P OUTPUT ACCEPT
```

a good idea ? If someone has other iptables rules ont their server, then they'll all be wiped.

DrXala:
Yes it's a good idea.
If we stop the iptables service, it's because we don't want any more rules. This service is general, not only for openvpn.

angristan:
I didn't realize it was in the service file. It makes sense indeed

angristan:
Works on : (install + reboot)

- CentOS 7
- Ubuntu 16.04
- Debian 8
- Debian 9

Works on Ubuntu 14.04, but not after a reboot. I suppose this is because there is no systemd ^^

Doesn't work on:

- Ubuntu 17.04/10 + Vultr:
```
Bad argument `ens3'
Try `iptables -h' or 'iptables --help' for more information.
```
It connects but the trafic is not forwarded.

angristan:
I think I found the reason. Idk if it's only on Vultr : 

```
root@vultr:~# ip -4 route ls | grep default | grep -Po '(?<=dev )(\S+)'
ens3
ens3
root@vultr:~# NIC=$(ip -4 route ls | grep default | grep -Po '(?<=dev )(\S+)')
root@vultr:~# iptables -t nat -A POSTROUTING -o $NIC -s 10.8.0.0/24 -j MASQUERADE
Bad argument `ens3'
Try `iptables -h' or 'iptables --help' for more information.
root@vultr:~#
```

angristan:
Ok, so that fixes it on Ubuntu 17.04. However, on Ubuntu 17.10 the trafic is still not routed, but I checked with master and the issue is also present...

angristan:
Sorry for the mess with the conflicts. But thanks for the PR!

PR #84, opened 2017-07-22
fix #8 Client files not beeing created in the right folder when using…

### LINKED ISSUE #8: Client files not beeing created in the right folder when using sudo
As reported here : https://github.com/Angristan/OpenVPN-install/issues/3.

When a normal user run the script using sudo, the .ovpn client conf is created in /root/, until you use `sudo -E`.

I'll have to find another way to get the home directory of the user.

Patlol:


angristan:
Hi, thanks.

What is `SUDO_USER` ? You seem to use it as a command but it does not exist for me. (Debian 8)

Patlol:
SUDO_USER is environment variable

try  
```
patrick@debian-VB:$ echo -e '#!/bin/sh\n echo $SUDO_USER' >test.sh
patrick@debian-VB:$ chmod +x test.sh
patrick@debian-VB:$ ./test.sh

patrick@debian-VB:$ sudo ./test.sh
patrick
patrick@debian-VB:$
patrick@debian-VB:$ sudo echo $SUDO_USER

patrick@debian-VB:$ 
```
give the user name using sudo  in script but  
give nothing in command line... :)

angristan:
Yes sorry >-< I meant variable. It is indeed not defined in the shell. Also `~` was referring to /home and not /root during the script, but you fixed it so it's ok.

Tested and approved :)

PR #91, opened 2017-08-21
Will now set "local" in server.conf to the chosen IP adderess

clickworkorange:
If you want to run OpenVPN in UDP mode on an secondary IP, UDP routing will fail unless you explicitly bind OpenVPN to the chosen IP address. This change includes the "local" parameter in the config and sets it to the IP address entered at the beginning.


angristan:
Thanks

Issue #9, opened 2016-12-08
Add user authentication for the private key


angristan:
Currently the script generates an all-in-one client configuration and does not ask for a password to decrypt the private key.

As this is a hardened OpenVPN install script, it would be nice a add an option to use a password when connecting to the VPN, to avoid anyone with the .ovpn file to be able to connect.

The modifications will have to be done [around here](https://github.com/Angristan/OpenVPN-install/blob/master/openvpn-install.sh#L409) with easy-rsa.

ohunecker:
First, you have to expand the [menu for add or remove a user cert](https://github.com/Angristan/OpenVPN-install/blob/master/openvpn-install.sh#L88) with the following option:
`2) 
			echo ""
			echo "Tell me a name for the client cert"
			echo "Please, use one word only, no special characters"
			read -p "Client name: " -e -i client CLIENT
			cd /etc/openvpn/easy-rsa/
			./easyrsa build-client-full $CLIENT
			# Generates the custom client.ovpn
			newclient "$CLIENT"
			echo ""
			echo "Client $CLIENT added, certs available at ~/$CLIENT.ovpn"
			exit
			;;`

After that you can change the [default creation of a user cer](https://github.com/Angristan/OpenVPN-install/blob/master/openvpn-install.sh#L409)t with this:
`while :
	do
	clear
		echo "What type of cert do you want to create?"
		echo "   1) Create a cert without password for a new user"
		echo "   2) Create a cert with password for a new user"
		read -p "Select an option [1-2]: " option
		case $option in
			1) 
			./easyrsa build-client-full $CLIENT nopass
			exit
			;;
			2) 
			./easyrsa build-client-full $CLIENT
			exit
			;;
		esac
	done`

rhbkweb:
On Nyr original script in order to generate a user key/cert with password he said that: "Just look at lines 84 and 217. If you remove nopass, it will simply ask for one during creation"

Does anyone knows what are the exact lines on this script?

angristan:
Look at my post :)

rhbkweb:
@Angristan 

From looking on your post it seems I must change the following 3 lines, correct?

409	./easyrsa build-server-full server nopass
410	./easyrsa build-client-full $CLIENT nopass
411	./easyrsa gen-crl

If yes, should I just remove the "nopass" on line 409 and 410 or further changes are needed? Thanks

angristan:
Removing nopass from lines 409 and 410 should do it

[...]

PR #92, opened 2017-08-27
Adding auth-nocache Option to .ovpn Configuration

LopezNathan:
Not necessarily required, but would like to add this to suppress the warning that is currently being produced when connecting.


x0r2d2:
@Angristan 
I think this PR has to be merged. :)

angristan:
Indeed. I'll merge this for now, but I'll add it dynamically when implementing https://github.com/Angristan/OpenVPN-install/issues/9

PR #108, opened 2017-10-02
Fix wrong deletion of ovpn configs

ghost:


angristan:
I wonder how I could have leaved that, but thanks!

Issue #78, opened 2017-07-03
SSL routines:tls_process_client_certificate:certificate verify failed


openmotion:
hello
i cannot connect anymore to my vpn server (archlinux)
i have this message in the log

```
192.168.0.1:37072 VERIFY ERROR: depth=0, error=CRL has expired: CN=client
192.168.0.1:37072 OpenSSL: error:1417C086:SSL routines:tls_process_client_certificate:certificate verify failed
```

so i changed in openssl.cnf this values :
default_days   = 3650         # how long to certify for
default_crl_days= 3650         # how long before next CRL

but i am not sure where are the certificates to execute the renew command : 
`openssl ca  -gencrl -keyfile keys/ca.key -cert keys/ca.crt  -out keys/crl.pem -config ./openssl.cnf`

thnaks

angristan:
Fixed by #109

Issue #82, opened 2017-07-16
Renew expired server certificate


openmotion:
hello
how can i renew the certificate of the server, because it's expired
thanks

openmotion:
cd /etc/openvpn/easy-rsa
./easyrsa gen-crl
cp /etc/openvpn/easy-rsa/pki/crl.pem /etc/openvpn/crl.pem
and start server

angristan:
Fixed by #109

PR #109, opened 2017-10-02
Extended the expiration date of the certificate revocation list to 10 years

jellemdekker:
Extended the expiration date of the certificate revocation list to 10 years. This fixes issues #78 and #82.


angristan:
Thanks

PR #134, opened 2017-12-01
Fedora Support

ardadem:
Finally, im pretty sure this ones works great on Fedora 27.


angristan:
Thanks!

What is your reason for adding `systemctl mask firewalld`?

ardadem:
However, firewalld always starting at boot (but disabled). So, i masked firewalld and solved.

angristan:
I confirm it works on KVM x64 with Fedora 25, 26 and 27. Thanks for the PR

PR #152, opened 2018-01-24
Fixed wrong home directory on newclient()

emilio1625:
Fixed some errors on the script that keep the client and server from doing the TLS handshake.
Also, i can confirm this is working on Fedora 27 and Fedora 27 armhf, but notice that the curve secp256r1 is not supported, only secp256k1 in neither the x86_64 or the armhf versions


angristan:
Wow, thanks, I can totally see this be the cause of weird issues that users may have been reporting. Thanks.

About this:

> but notice that the curve secp256r1 is not supported, only secp256k1 in neither the x86_64 or the armhf versions

I understand secp256r1 isn't available for you on armhf, but what do you mean with 

> only secp256k1 in neither the x86_64 or the armhf versions

Thanks :)

Issue #159, opened 2018-02-12
Please apply two small patches


jnko:
Please apply my two small patches, 

1st: Added option to create a password protected client key
2nd: Added support for Debian 9 (stretch)

Sorry for providing this way, but those changes are way too small to fork..

Thanks for your effort

```
105c105,106
<               echo "   1) Add a cert for a new user"
---
>               echo "   1) Add a cert for a new user (nopass)"
>               echo "   2) Add a cert for a new user (with private-key password)"
124a126,138
>                       echo ""
>                       echo "Tell me a name for the client cert"
>                       echo "Please, use one word only, no special characters"
>                       read -p "Client name: " -e -i client CLIENT
>                       cd /etc/openvpn/easy-rsa/
>                       ./easyrsa build-client-full $CLIENT
>                       # Generates the custom client.ovpn
>                       newclient "$CLIENT"
>                       echo ""
>                       echo "Client $CLIENT added, certs available at $homeDir/$CLIENT.ovpn"
>                       exit
>                       ;;
>                       3)
154c168
<                       3)
---
>                       4)
201c215
<                       4) exit;;
---
>                       5) exit;;
339a354,359
>                       wget -O - https://swupdate.openvpn.net/repos/repo-public.gpg | apt-key add -
>                       apt update
>               fi
>               # Debian 9
>               if [[ "$VERSION_ID" = 'VERSION_ID="9"' ]]; then
>                       echo "deb http://build.openvpn.net/debian/openvpn/stable stretch main" > /etc/apt/sources.list.d/openvpn.list
```

Kcchouette:
For the 2nd patch: the commit https://github.com/Angristan/OpenVPN-install/commit/cd013295851e98b3f784cc504dab4d36e15bb6df added the support for debian 9

PR #160, opened 2018-02-12
Add password option for clients

angristan:
This PR adds the option to add a password to the client file. It will be asked when connecting to the OpenVPN server. Basically, this encrypts the client's private key with a password.

Using a password for your private key is more secure, as not having one mean anyone with your .ovpn file can connect to your server and/or decrypt your traffic if they manage to make a MitM.

The user will be given the option when installing the server for the first time (and thus creating the first user) and when adding new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ This fork includes the following features :
 - [Arch Linux support](https://github.com/Angristan/OpenVPN-install/pull/2)
 - Up-to-date OpenVPN thanks to [EPEL](http://fedoraproject.org/wiki/EPEL) for CentOS and [swupdate.openvpn.net](https://community.openvpn.net/openvpn/wiki/OpenvpnSoftwareRepos) for Ubuntu and Debian. These are third-party yet trusted repositories.
 - Randomized certificate name
+- The ability to create passwordless clients and clients protected with a password
 - Other improvements !
 
 ## DNS
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -127,7 +127,7 @@ if [[ -e /etc/openvpn/server.conf ]]; then
 				./easyrsa build-client-full $CLIENT nopass
 				;;
 				2)
-				echo "⚠️ You will be asked for the client password below"
+				echo "⚠️ You will be asked for the client password below ⚠️"
 				./easyrsa build-client-full $CLIENT
 				;;
 			esac
@@ -508,7 +508,7 @@ WantedBy=multi-user.target" > /etc/systemd/system/iptables.service
 			./easyrsa build-client-full $CLIENT nopass
 		;;
 		2)
-			echo "⚠️ You will be asked for the client password below"
+			echo "⚠️ You will be asked for the client password below ⚠️"
 			./easyrsa build-client-full $CLIENT
 		;;
 	esac
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -122,6 +122,7 @@ if [[ -e /etc/openvpn/server.conf ]]; then
 			cd /etc/openvpn/easy-rsa/
 			case $pass in
 				1)
+				echo "⚠️ You will be asked for the client password below"
 				./easyrsa build-client-full $CLIENT nopass
 				;;
 				2)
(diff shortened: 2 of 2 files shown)


angristan:
Fixes #9 and #159

angristan:
So this is working fine, but atm the password is asked when `./easyrsa` is called. I would like to ask the user for its password first, store it in a variable to use it with easyrsa laster.

angristan:
I was thinking about add `./easyrsa build-client-full $CLIENT <<< $password` but easy-rsa generates the key before asking for the password so it's not possible...

jellemdekker:
_General solution 1: pipes_ in [this Stackoverflow answer](https://stackoverflow.com/a/22422834) might offer some guidance.

cezar97:
Spawn, expect and send may be good options too.
https://askubuntu.com/a/307080

PakTam:
it is possible to create Multiple client using single KEY?
as this website https://sshdropbear.net/

so then the client can be add using WEBMIN or OCS PANEL

cezar97:
@PakTam yes it's possible but this script just doesn't do it because it isn't recommended from a security point of view and this way you can also revoke single users, too.
Anyway I think it's the same certificate rather than the same key.

cezar97:
> So this is working fine, but atm the password is asked when ./easyrsa is called. I would like to ask the user for its password first, store it in a variable to use it with easyrsa later.

Isn't pretty much the same thing writing the pass directly when easyrsa ask so or store it in a variable and use it later for the same thing? What benefits are there in storing it as a variable instead?

angristan:
It's exactly the same except it is more clear to ask for a client password and put it in a variable than having a random prompt in the middle of easy-rsa stuff waiting for something.

heikoh81:
I just installed your script for the first time, thanks!
For use on my mobile phones I'd also prefer to have a scripted Option to ask for a Password.
I figured out how to Change the .sh-file as suggested by craig_mld in #9 
You could say "leave empty if no Password is wanted", but then, it would be easy to add one if needed.

fritzmg:
👍 for this :). Is there any reasony why this isn't merged yet to master?

angristan:
It currently asks the password here:

screenshot_26-05-2018_20-26-08

Does it look acceptable to you?

fritzmg:
Oh I see. Well for me it's good enough, but I understand that it would be more user friendly if it was asked in the beginning.

heikoh81:
When I first used your script about 1-2 month ago, I had to modify the script file as suggested in #9 (remove NOPASS in the code):
https://github.com/Angristan/OpenVPN-install/issues/9
So I now have 2 scripts that I uses: 1 for creating users without password (Raspis) and 1 for users with Passwords (mobile use).

This is very complicated, and thus I suggested you integrate it in your script.
I think the script should state if you leave password empty,  the user is created without password.
I don't see this in the screenshots attached. It just seems to offer the possibility to enter a password - but what happens if none is entered?

fritzmg:
> I don't see this in the screenshots attached. It just seems to offer the possibility to enter a password - but what happens if none is entered?

You can choose whether you want to use a password or not.

heikoh81:
Where is that option? Can't see it in the screenshot?
If I just leave the password blank than there is no password?
(Don't have a Test-system running to test the new version...)

cezar97:
@heikoh81 generally if you put no password then it's generated without one. This script ask if you want a password or not so there's no need to leave it empty when not used.

heikoh81:
Perfect. So that feature I implemented now and no need any more to edit the code yourself?

cezar97:
Well, only when this PR will be merged. Or if you want to speed up things, clone / fork the repo and merge this branch manually into the master. Don't use this branch alone as without merging them both you lose all the new commits on master. @heikoh81

heikoh81:
So let's hope it is merged.
I'm just an advanced enduser, no developer on Github...

cezar97:
@Angristan Instead of leaving PRs open for a long time, especially with this one that has a feature requested by a lot of users, you can `echo "When asked for PEM pass prashe write your password"` so all users know what to do and just merge it. Later you can update it with variables and other suff.
Even if doesn't look acceptable to me either.

newsera:
Can we have this feature implemented for 2.4-update as well?

PR #96, opened 2017-09-14
OpenVPN 2.4 support

angristan:
This is WIP for adding OpenVPN 2.4 new features into the script

- Add support for AES-GCM ciphers for the data channel
- Add support for tls-crypt
- Add support for ECDSA certificates
- Add support for ECDHE
- Add choice for HMAC auth algorithm
- Add choice for certificate hash algorithm
- Add choice for the control channel's cipher

All these options have an OpenVPN 2.3-compatible choice (example : RSA cert and DH key)

In my previous PR, I removed the encryptions choices to only choose the best ones. Here I've added the choice for every encryption parameter, as it seems that users prefer to use the script this way.

It seems to be working, I'll have to dome some more testing, rewrite some menus and then rewrite the readme with to add the new features.

PR #163, opened 2018-02-14
Remove unsupported version of ubuntu

Kcchouette:
Remove 12.04 as the support finished on April 28, 2017
Remove 16.10 as the support finished 2017-07-20
Remove 17.04 as the support finished 2018-01-13


angristan:
Already done in #96, but we can push this in master now

PR #187, opened 2018-03-29
Improve Debian/Ubuntu version checking

xiagw:
Here is a better way or a more standard generic method

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -117,6 +117,8 @@ if [[ -e /etc/openvpn/server.conf ]]; then
 		case $option in
 			1)
 			echo ""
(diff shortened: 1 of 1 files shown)


openvpn-install.sh
```diff
@@ -110,6 +117,8 @@ if [[ -e /etc/openvpn/server.conf ]]; then
 		case $option in
 			1)
 			echo ""
+			echo "Here are the files that already exist,do not repeat that"
```
jellemdekker: What do these two lines do? How do they help in determining the OS version better?

angristan:
I reworked it do make it simpler.

angristan:
Thanks though 👍

PR #196, opened 2018-04-05
Avoid Arch Linux & Fedora workaround

yapbreak:
Replace all occurence of /etc/openvpn by OVPN_ROOT variable.
Set this variable accordingly after OS detection to
        - /etc/openvpn/server on Arch and Fedora
        - /etc/openpvn to all other distros


angristan:
Breaks the service on Fedora 28.

letroll:
this helped me on archlinux but I have failed on first try and was successful after removed the server.conf file and created manually the path : /etc/openvpn/server/easy-rsa/ and the systemd file and retry

PR #198, opened 2018-04-14
Plaintext cert in ovpn not needed

ghost:


angristan:
But it could be useful in order get the algorithm or expiration date without using OpenSSL 🤔

cezar97:
I've thought this too, but then I've remembered that there are still the connections logs that shows certs details like the algorithm. For the expiration part, I don't really care as there's enough time.
The PR is because the ovpn looks pretty bad with a pasted plaintext cert in the middle.

Angristan:
I agree, but let's keep following the RFCs as these fields are not optional (even though the are technically)

https://en.wikipedia.org/wiki/X.509#Structure_of_a_certificate

cezar97:
I mean for an end user the certificate details shows in the logs, while the expiration can still be viewed on server side with a `cat` on whatever certificate issued by the sys-admin: no need for openssl.
I didn't get the RFC part, as there is still the certificate in PEM format, just not the plaintext one. It connects anyway, so it's only a "styling" change in the ovpns...

angristan:
I can't resolve conflicts since the original branch does not exist. I added it here: https://github.com/angristan/openvpn-install/commit/8d5d080cc06e10329968b5bd8e85d165fe8321af

Thanks!

PR #208, opened 2018-05-01
Custom DNS support and better codes

sayem314:
I have added support for custom DNS and improved overall coding. Also fixed a DNS(9) bug.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -243,9 +243,19 @@ else
 	echo "   7) Google (Anycast: worldwide)"
 	echo "   8) Yandex Basic (Russia)"
 	echo "   9) AdGuard DNS (Russia)"
-	until [[ "$DNS" -gt 0 && "$DNS" -le 9 && "$DNS" =~ ^[0-9]$ ]]; do
+	echo "   10) Custom"
+	until [[ "$DNS" -gt 0 && "$DNS" -le 10 && "$DNS" =~ ^[0-9]$ ]]; do
 		read -rp "DNS [1-8]: " -e -i 1 DNS
 	done
+	if [[ $DNS = "10" ]]; then
+		# Get DNS IP and validate
+		until [[ "$DNS1" =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$ ]]; do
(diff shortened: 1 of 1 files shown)


angristan:
Hello, and thank you for your PR.

Could you please explain what does `read -rp` enhances? What about `|| return`?

Also if it doesn't bother you, could you do 2 separate PRs? 🙏

jellemdekker:
1. > Could you please explain what does read -rp enhances?

From the [bash manpage](http://linuxcommand.org/lc3_man_pages/readh.html) about the built-in `read` command:
> -p prompt:	output the string PROMPT without a trailing newline before
    		attempting to read
      -r:		do not allow backslashes to escape any characters



 
2. > What about || return?

From [shellcheck's wiki](https://github.com/koalaman/shellcheck/wiki/SC2164):
> Rationale:
cd can fail for a variety of reasons: misspelled paths, missing directories, missing permissions, broken symlinks and more.
> 
> If/when it does, the script will keep going and do all its operations in the wrong directory. This can be messy, especially if the operations involve creating or deleting a lot of files.
> 
> To avoid this, make sure you handle the cases when cd fails. Ways to do this include
> 
> **cd foo || exit as suggested to just abort immediately**
> if cd foo; then echo "Ok"; else echo "Fail"; fi for custom handling
> <(cd foo && cmd) as an alternative to <(cd foo || exit; cmd) in <(..), $(..) or ( )

Except instead of `cd foo || exit`, it's now `cd foo || return`.

jellemdekker:
@sayem314 Please split up your pull request so that the code for support for custom DNS is in a separate PR.

angristan (approved):

sayem314:
I’m on vacation for Ramadan. Can’t really make pull request now. I have to update it after Eid. In the meantime if anyone would like to work on that I have no problem with that 🙂 And sorry about real late reply 😔

sayem314:
Closing this in favor of new PR

angristan:
But custom DNS is gone D:

sayem314:
Oh yeah. Forgot about that. Also I broke my wrist 😔

Anyway I will make sure to pull request for custom DNS shortly.

PR #215, opened 2018-05-09
TCP_NODELAY

x0r2d2:
Add tcp_nodelay option for proto tcp

 --tcp-nodelay
              This  macro sets the TCP_NODELAY socket flag on the server as well as pushes it to
              connecting clients.  The TCP_NODELAY flag disables  the  Nagle  algorithm  on  TCP
              sockets  causing  packets  to  be transmitted immediately with low latency, rather
              than waiting a short period of time in order to aggregate several packets  into  a
              larger  containing packet.  In VPN applications over TCP, TCP_NODELAY is generally
              a good latency optimization.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -690,7 +690,6 @@ WantedBy=multi-user.target" > /etc/systemd/system/iptables.service
 	elif [[ "$PROTOCOL" = '2' ]]; then
 		echo "proto tcp" >> /etc/openvpn/server.conf
 		echo "socket-flags TCP_NODELAY" >> /etc/openvpn/server.conf
-		echo "push "socket-flags TCP_NODELAY"" >> /etc/openvpn/server.conf
 	fi
 	echo "dev tun
 user nobody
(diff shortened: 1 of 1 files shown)


cezar97:
It should be escaped with echo 'push "socket ... " '.

x0r2d2:
@cezar97 
It will be better to add option to client configs?

cezar97:
Well, supposing that Angristan agrees with the support of this, I think the option on client side has as the advantage the ability to comment it when not needed.

angristan:
If it works when you enable it only from the client-side, then yes I agree to only put it there

cezar97:
@Angristan from the man page:
> This option is pushable from server to client, and should be used on both client and server for maximum effect.

So it can be used on client side only.

angristan:
What does "maximum effect" even mean lol

jellemdekker:
@hybtoy Please update your PR so it can be merged.

angristan:
After more thinking, since this option is available since at least OpenVPN 2.1 and does not seem to have drawbacks, I agree to put it on both sides. Let me just test this branch and I will merge it.

angristan:
Is there a reason you want to marge into `openvpn-2.4-update` and not `master`?

x0r2d2:
@Angristan there is no reason. Can be merged to master.

x0r2d2:
@jellemdekker what I have to change? Branch?

jellemdekker:
> @jellemdekker what I have to change? Branch?

Never mind, you can ignore my comment.

cezar97:
@Angristan why still 2.4 branch?
Also, someone did performance / benchmark tests?

angristan:
I get 

```
NOTE: setsockopt TCP_NODELAY=1 failed
```

on both sides.

I'm closing this until a proper tested PR is made.

PR #228, opened 2018-05-28
[FIX] Unable to select AdGuard DNS choice

Jebtrix:


sayem314:
@Angristan The label should be “bugfix”

angristan:
thanks @sayem314

PR #229, opened 2018-05-28
Add option to generate random port in private port range

Jebtrix:
I always use a random non default port so why not let the script make it convenient.


ghost (changes requested):
Clean PR

PR #230, opened 2018-05-29
 [ENHANCEMENT] Server log level option 

Jebtrix:
Just another option I find my self changing manually. I'm sure there are other users that would just like to disable logging completely after verifying server works on a particular host OS/VPS.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -242,8 +242,8 @@ else
 	echo "   7) Google (Anycast: worldwide)"
 	echo "   8) Yandex Basic (Russia)"
 	echo "   9) AdGuard DNS (Russia)"
-	while [[ $DNS != "1" && $DNS != "2" && $DNS != "3" && $DNS != "4" && $DNS != "5" && $DNS != "6" && $DNS != "7" && $DNS != "8" ]]; do
-		read -p "DNS [1-8]: " -e -i 1 DNS
+	while [[ $DNS != "1" && $DNS != "2" && $DNS != "3" && $DNS != "4" && $DNS != "5" && $DNS != "6" && $DNS != "7" && $DNS != "8" && $DNS != "9" ]]; do
+		read -p "DNS [1-9]: " -e -i 1 DNS
 	done
 	echo ""
 	echo "See https://github.com/Angristan/OpenVPN-install#encryption to learn more about "
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -326,6 +326,15 @@ else
 		;;
 	esac
 	echo ""
+	echo "Choose server log level:"
+	echo "   1) Normal (verbose level 3)"
+	echo "   2) High (verbose level 4)"
+	echo "   3) Debug (verbose level 6)"	
+	echo "   4) Off (verbose level 0, logs to /dev/null)"
+	while [[ $SERVER_LOG_LEVEL != "1" && $SERVER_LOG_LEVEL != "2" && $SERVER_LOG_LEVEL != "3" && $SERVER_LOG_LEVEL != "4" ]]; do
+		read -p "Server log level [1-4]: " -e -i 1 SERVER_LOG_LEVEL
+	done
+	echo ""
 	echo "Finally, tell me a name for the client certificate and configuration"
 	while [[ $CLIENT = "" ]]; do
 		echo "Please, use one word only, no special characters"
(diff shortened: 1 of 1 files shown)


angristan:
This could have been great, but the log levels are only useful for the clients, it doesn't change anything on the server, afaik.

cezar97:
I think it's not so nice to see this option. It should be only necessary stuff so it's more straightforward. Not saying it's not useful... But as it's for advanced users, the script can look for an environmental variable instead.
As Angristan said, the `status` is the one that matters. So in `off mode` you can still leave `verb x`.

angristan:
Closing this. Log verbosity has to be configured on the client side.

perfectpeach (commented):
Hiya @Jebtrix



 Is this for disabling al logs that Openvpn installed by @angristan automated script disable all logs?
If yes, why this is closed and how shall I disable all logs myself manually?


Tnx and best of luck

randshell:
@perfectpeach check `server.conf` to have this:
```
log /dev/null
status /dev/null
verb 0
```
or to disable all logs on client change `client.ovpn` `verb` option to `verb 0`.

perfectpeach:
> @perfectpeach check `server.conf` to have this:
> 
> ```
> log /dev/null
> status /dev/null
> verb 0
> ```
> 
> or to disable all logs on client change `client.ovpn` `verb` option to `verb 0`.

Hiya @randomshell 



Thanks. I do both settings on my VPS based OpenVPN. 

Q) But if I do the client side 'verb 0' on client file of client.ovpn on a commercial OpenVPN VPN service provider, does this will block the commercial OpenVPN VPN service provider to logging my IP as well or if I do this on commercial OpenVPN VPN service provider, this will not work for blocking them from logging my actual ISP IP address on the OpenVPN VPN service provider too?

Tnx and best of luck

perfectpeach:
Hiya @randomshell 



Thanks.  Q 1) How to find and delete already created logs? 

I'm running a CentOS7.5-x64 on VPS, I installed OpenVPN community edition on it by this script.

I want to disable all logs and delete everything that is logged, anything that can trace back to me and reveal my IP. How shall I disable all logs and then where are the name of current logs that I should delete?

e.g. I see the file named 'status.log' in folder /var/log/openvpn/ have my actual IP address, I don't want to this be created on the server and I want to disable all logs and delete all logs. How shall I do this?

/var/log/openvpn/ 
contain below:

. .. status.log .status.log.swo


2) Is there any logs in

/var/log/syslog
As I don't see any 'syslog' in /var/log/ folder, there is only below in,

/var/log/
. boot.log-20200304 dmesg.old lfd.log secure yum.log

.. btmp firewalld maillog spooler
anaconda chrony grubby messages tallylog
audit cron grubby_prune_debug openvpn tuned
boot.log dmesg lastlog rhsm wtmp

How shall I disable all logs that is related to OpenVPN and its connections and IP's here, if exists? And how to delete all logs that already is created?


3) Where is the file below,

openvpn.log
as I don't see this? How to disable this log and how to find and delete this file?


4) The directory below

/etc/openvpn/
contain the followings, is there any logs that need to be disabled here? Where can I delete the already created logs here or by files here:

. client ipp.txt server.conf
.. client-template.txt server tls-crypt.key
ca.crt crl.pem server_{alphanumeric sequence removed by this text to prevent tracking}.crt
ca.key easy-rsa server_{alphanumeric sequence removed by this text to prevent tracking}.key


5) On servers, OpenVPN is usually run as a system service, i.e., started with the --daemon option. According to the OpenVPN manual page, using the --daemon [progname]
option has the following effect:

Become a daemon after all initialization functions are completed. This option will cause all message and error output to be sent to the syslog file (such as /var/log/messages
), except for the output of scripts and ifconfig commands, which will go to /dev/null
unless otherwise redirected. The syslog redirection occurs immediately at the point that --daemon
is parsed on the command line even though the daemonization point occurs later. If one of the --log
options is present, it will supercede (sic) syslog redirection.

Thus there may be logs in system logs that is related to the OpenVPN, how shall I find and disable these logs in CentOS7.5-x64? Where is the already logged files and how shall I find and delete them?


6) Is there any of the ' *.conf ' files including ' server.conf ' or ' <client name/>.conf ' have any option to disable the logs, where are they located and how to disable logs in them and where are the already created logs that should get deleted?


7) Is the 'openvpn' as a service that is running, will create any logs and if so, how shall I disable these logs it and then how shall I delete already created logs?



Tnx and best of luck

randshell:
If this is what really concerns you remember that your vps provider has other ways to see what you do that doesn't necessarily means looking at the logs you want to delete. They can use Wireshark for example too look at your traffic and you can't know it :smile:. Just use a provider you trust and don't worry too much about it. Or use Tor if you think is better for you but I consider a vpn to be enough for most users.
See [this](https://www.privacytools.io/providers/vpn/) for a hosted vpn or [this](https://www.privacytools.io/providers/hosting/) for a provider if you want to host it. (I'm not affiliated with the website linked)

For the record, the logs openvpn creates are connection logs, it doesn't save what websites you visit etc. so there is little point in deleting them.

Issue #236, opened 2018-06-05
IPv6 NAT support


galaxydata:
```
echo "net.ipv6.conf.all.forwarding = 1" >> /etc/sysctl.conf
sysctl -p

```
add
/etc/openvpn/server.conf

```
#IPv6 config
server-ipv6 fd6c:62d9:eb8c::/112
proto udp6
tun-ipv6
push tun-ipv6
push "route-ipv6 2000::/3"
push "redirect-gateway ipv6"
```

`systemctl restart openvpn@server.service`

ADD IP6TABLES
rm /etc/sysconfig/ip6tables

edit
`nano /etc/sysconfig/ip6tables`

```
ADD 
*filter
:INPUT ACCEPT [0:0]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
:XL-Firewall-1-INPUT - [0:0]
-A INPUT -j XL-Firewall-1-INPUT
-A FORWARD -j XL-Firewall-1-INPUT
-A XL-Firewall-1-INPUT -i lo -j ACCEPT
-A XL-Firewall-1-INPUT -p icmpv6 -j ACCEPT
-A XL-Firewall-1-INPUT -m state --state ESTABLISHED,RELATED -j ACCEPT
-A XL-Firewall-1-INPUT -m state --state NEW -m udp -p udp --dport 1194 -j ACCEPT
-A XL-Firewall-1-INPUT -m state --state NEW -m tcp -p tcp --dport 80 -j ACCEPT
-A XL-Firewall-1-INPUT -i tun+ -j ACCEPT
-A XL-Firewall-1-INPUT -j REJECT --reject-with icmp6-adm-prohibited
COMMIT
*mangle
:PREROUTING ACCEPT [0:0]
:INPUT ACCEPT [0:0]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
:POSTROUTING ACCEPT [0:0]
COMMIT
*nat
:PREROUTING ACCEPT [0:0]
:POSTROUTING ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
-A POSTROUTING -s fd6c:62d9:eb8c::/112 -j MASQUERADE
COMMIT
```

`systemctl restart ip6tables`

angristan:
Thanks a lot :) 

We'll probably get murdered by implementing NATed IPv6 but that's better than nothing I guess

angristan:
See #238

PR #238, opened 2018-06-06
Add IPv6 NAT support

angristan:
Thanks to @galaxydata (#236)

**This isn't *real* IPv6 support**, but since it's easy to implement, why not support it?

To do:

- [x] Handle all the firewall stuff
- [x] Detect IPv6 connection on the host before asking
- [x] Choose a cool subnet
- [x] Test on multiple OS and providers
- [x] Update README

Note the connection is made via IPv4 and IPv6 settings are pushed **from** the server


angristan:
@cezar97 We don't have too, it will work without them, but yes we should do that!

7twin:
@Angristan this worked just fine on Ubuntu 16.04LTS (the original ipv6 is not leaked anymore)

cantalupo555:
@Angristan Finally support IPv6, thank you!

ookangzheng:
I choose TCP, but the conf file TCP doesn't give me IPV6 version
should be 

```
// TCP ipv6
proto tcp6-server
```
Second,
in the conf file IPv6 part,
It still gives me proto udp6 

Other work fine

angristan:
Ok, it's working.

`ip6tables` won't be applied at boot but I will rework the firewall management after this.

PR #249, opened 2018-07-06
Fix UFW default forwarding policy on Ubuntu

angristan:
On Ubuntu, the `DEFAULT_FORWARD_POLICY` in `/etc/default/ufw` is DROP, so it can affect OpenVPN.


angristan:
Not needed since https://github.com/angristan/openvpn-install/pull/291

PR #271, opened 2018-08-18
Refactoring 🛠(Cancelled)

angristan:
The script is over 700 lines right now and it's a pain to maintain.

The goal of this PR is to break the script into functions and somewhat refactor everything to make it simpler and easier to understand.


angristan:
Too much conflicts, I'll start from scratch

PR #288, opened 2018-09-14
Add self-hosted DNS resolver (Unbound)

### LINKED ISSUE #5: Self-hosted DNS Resolver (Unbound)
As discussed in https://github.com/Angristan/OpenVPN-install/issues/4, having the option to use a local DNS server would be great for the following reasons : 

- It will be fast all aroung the world, as it will allow a local cache and will communicate directly with the root servers
- It's neutral
- Logs will be disabled by default, and if they are enabled, they'll be only accessible on the server
- It does not depend on any third-party server, thus ensuring 100% availability and neutrality

I already made a script to install a local DNS server on a Linux machine : https://github.com/Angristan/Local-DNS-resolver

I will use unbound, and adapt the configuration to allow queries from the private network (10.8.0.0/24), and then use 10.8.0.1 as the DNS resolver in the OpenVPN configuration.

~If unbound is already installed on the server where the script is ran, I think I'll ask for a confirmation to overwrite `/etc/ubound.conf`, and if the answer is no, ask the user to select other DNS servers.~

Any suggestion is welcomed until I implement the feature.

randshell:
Added option for installing and configuring Unbound as a recursive resolver. Fix #5.

Notes:
- Option added as a function for easier integration when refactoring branch is merged.
- Changes are a backport of @angristan local-dns-resolver repository.
- ~~Handler when Unbound is already installed is WIP.~~
- I didn't do any test.
- Feedback and ideas are appreciated.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -355,7 +355,7 @@ if [[ -e /etc/openvpn/server.conf ]]; then
 
 					rm /etc/unbound/${UNBOUND_CONF_D}/openvpn.conf
 
-					read -rp "Do you want to remove Unbound, too? [y/n]: " -e -i n REMOVE
+					read -rp "Do you want to remove Unbound, too? [y/n]: " -e REMOVE
 
 					if [[ "$REMOVE" = 'y' ]]; then
 						if [[ "$OS" = 'debian' ]]; then
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -805,9 +805,6 @@ prefetch: yes' >> /etc/unbound/unbound.conf
 			# Restart Unbound
 			service unbound restart
 
-			# Needed for the chattr command
-			apt-get install -y e2fsprogs
-
 		elif [[ "$OS" = "centos" ]]; then
 			# Install Unbound
 			yum install -y unbound
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -189,9 +189,19 @@ private-address: ::ffff:0:0/96" >> /etc/unbound/unbound.conf
 			if [ -d "/etc/unbound/conf.d" ]; then
 				# Debian / Ubuntu
 				UNBOUND_CONF_D="/etc/unbound/conf.d"
+				if ! grep -qs "^include: \"/etc/unbound/unbound.conf.d/\*.conf\"" /etc/unbound/unbound.conf; then
+					echo 'include: "/etc/unbound/unbound.conf.d/*.conf"' >> /etc/unbound/unbound.conf
+				fi
 			elif [ -d "/etc/unbound/unbound.conf.d" ]; then
 				# CentOS / Fedora
 				UNBOUND_CONF_D="/etc/unbound/unbound.conf.d"
+				if ! grep -qs "^include: /etc/unbound/conf.d/\*.conf" /etc/unbound/unbound.conf; then
+					echo 'include: /etc/unbound/conf.d/*.conf' >> /etc/unbound/unbound.conf
+				fi
+			else
+				UNBOUND_CONF_D="/etc/unbound/conf.d"
+				mkdir -p $UNBOUND_CONF_D
+				echo 'include: "/etc/unbound/unbound.conf.d/*.conf"' >> /etc/unbound/unbound.conf
 			fi
 
 			# Add OpenVPN integration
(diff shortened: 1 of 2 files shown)


randshell:
I think the only thing left is testing this integration. Unfortunately, I only have a slow machine currently, so no VMs where I can test :(

openvpn-install.sh
```diff
@@ -780,3 +809,106 @@ verb 3" >> /etc/openvpn/client-template.txt
...
+		echo "an `include:` statement will be added to `unbound.conf` with the necessary changes in a separate `openvpn-server.conf` file."
+		echo "No other changes are made to the current configuration."
+
+		while [[ $CONTINUE != "y" && $CONTINUE != "n" ]]; do
+			read -rp "Apply configuration changes? [y/n]: " -e local CONTINUE
+		done
+
+		if [[ $CONTINUE = "y" ]]; then
+			# Add include: statement
+			awk '{ print } !flag && /server:/ { print "        include: /etc/unbound/openvpn-server.conf"; flag = 1 }' /etc/unbound/unbound.conf > /etc/unbound/unbound.conf
```
angristan: Can you explain this line?

angristan:
So, what happens if there bind or something else using port 53? :)

openvpn-install.sh
randshell: It appends a line (`inlcude:... `) after another (`server:`), on the first occurrence only.

randshell:
@angristan It is configured to listen on 10.8.0.1, but this interface is new because the OpenVPN installation created it. There can't be any packages listening on it.
It can cause problems only if there's something on 0.0.0.0
Indeed, as mentioned, it needs some testing for tweaks like these.

Are you suggesting using the same `lsof -i :53`? In this case there's still the 0.0.0.0 problem, again. If I remember correctly, the only distro that fails this check is Ubuntu with its pre-installed package that run on port 53. Not sure about Debian.

angristan:
> @angristan It is configured to listen on 10.8.0.1, but this interface is new because the OpenVPN installation created it.

Indeed 👍

I will test with Ubuntu

randomshell:
Thanks @Angristan.

Regarding 99d4b59, is expat required on Arch, or it was only for chattr?

angristan:
- Debian *vanilla*: ok
- Ubuntu (with systemd-resolved on `127.0.0.53:53`: ok
- Debian with Unbound: nok

openvpn-install.sh
angristan: This is not working, is echoes an empty string to `/etc/unbound/unbound.conf`.

angristan: There is not `server:` by default on Debian

randshell: https://manpages.debian.org/stretch/unbound/unbound.conf.5.en.html
> Below is a minimal config file. The source distribution contains an extensive example.conf file with all the options.

Good docs, then.

angristan: I mean there is one, but not in `/etc/unbound/unbound.conf` directly.

```sh
root@vultr:~# cat /etc/unbound/unbound.conf
# Unbound configuration file for Debian.
#
# See the unbound.conf(5) man page.
#
# See /usr/share/doc/unbound/examples/unbound.conf for a commented
# reference config file.
#
# The following line includes additional configuration files from the
# /etc/unbound/unbound.conf.d directory.
include: "/etc/unbound/unbound.conf.d/*.conf"
```

```sh
root@vultr:~# cat /etc/unbound/unbound.conf.d/root-auto-trust-anchor-file.conf
server:
    # The following line will configure unbound to perform cryptographic
    # DNSSEC validation using the root trust anchor.
    auto-trust-anchor-file: "/var/lib/unbound/root.key"
```

randomshell:
https://github.com/angristan/openvpn-install/pull/288/commits/21734bab5fbb242aa94b334701d62326268bb8b6
Doesn't require an include: in unbound.conf?

openvpn-install.sh
```diff
@@ -355,7 +355,7 @@ if [[ -e /etc/openvpn/server.conf ]]; then
 
 					rm /etc/unbound/${UNBOUND_CONF_D}/openvpn.conf
 
-					read -rp "Do you want to remove Unbound, too? [y/n]: " -e -i n REMOVE
+					read -rp "Do you want to remove Unbound, too? [y/n]: " -e REMOVE
```
randshell: If there is no default value, we should use a while loop do avoid blank variable.

randshell: In this case, the variable name should be changed too because there is already a $REMOVE in the OpenVPN uninstall. Otherwise it skips the loop.

angristan:
Ok this is working as expected now, when unbound is already installed.

I will test on CentOS/Fedora before merging

PR #290, opened 2018-09-16
Update EasyRSA to 3.0.5

angristan:


angristan:
It's breaking something:

```xml
<ca>
</ca>
<cert>
</cert>
<key>
</key>
```

angristan:
The release archive is not yet available.

PR #291, opened 2018-09-16
Rework iptables handling

angristan:
The way we handled iptables rules was pretty messy. 

We switched from rc.local to systemd a while ago: https://github.com/angristan/openvpn-install/pull/83.

But it was still confusing and messed too much with the system since we flushed all the rules. Also it wasn't always working after a reboot, and everything wasn't removed after an uninstallation.

I reworked all of this. Now we have:

- A script that adds the needed rules
- A script that remove these rules
- A systemd service to manage these scripts
- During install and boot, the service executes the script that adds the rules
- During removal, the service executes the script that removes the rules, and then both are removed

We don't touch the rest of the rules nor the policies anymore. I made sure that the rules are working at boot and that everything we added is removed when removing OpenVPN.

**TL;DR:** It's simpler, unobtrusive, works after reboots, and is cleaned after removal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,6 @@ The script is made to work on these OS and architectures :
 
 - **Debian 8** (i386, amd64)
 - **Debian 9** (i386, amd64, armhf, arm64)
-- **Ubuntu 14.04 LTS** (i386, amd64)
 - **Ubuntu 16.04 LTS** (i386, amd64, armhf)
 - **Ubuntu 17.10** (i386, amd64, armhf, arm64)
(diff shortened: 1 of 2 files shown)


angristan:
I removed support for Ubuntu 14.04 since we need systemd. (We already needed it).

I tested with UFW enabled in Ubuntu 18.04, it should be working fine.

angristan:
Oh and IPv6 is supported now.

angristan:
Ok so I tested on all distributions, even with unbound. Everything is OK ✅

Issue #302, opened 2018-09-23
secp256r1 --> secp256k1


kimberct:
Curve labeled secp256r1 needs to be secp256k1.  At least on 2.4.6.

Nice job on the script.  Really glad to see the 2.4-update branch pulled in.

angristan:
Good catch. `secp256r1` is in fact `prime256v1`. `secp256k1` is a Koblitz curve,

angristan:
FTR:

- easy-rsa handles the misuse like this: `using curve name prime256v1 instead of secp256r1`
- OpenVPN: `Failed to use supplied curve (UNDEF), using secp384r1 instead.`

angristan:
Fixed in https://github.com/angristan/openvpn-install/commit/70ebe5620db5c0f9ca218d72832962b159397da7

iProdigy:
@Angristan I'd rather recommend k1 as it's less likely to be botched by the NSA

angristan:
NSA is not this script's threat model.

iProdigy:
Understood. Would you be open to adding the k1 option to this script (I don't mind making the PR)?

angristan:
I will do some tests.

iProdigy:
Also wanted to note that: if there does, in fact, exist some vulnerability in P-256, it has the potential to be stolen from the NSA, putting many users at risk when there are alternative ciphers with more sensible parameters.

angristan:
Let's talk again when there is, in fact, a vulnerability in P-256. :)

iProdigy:
Hah, alright - it's not implausible given the Dual_EC_DRBG fiasco and strange parameter choice.

angristan:
Elliptic Curves are kind of a mess... ^^

iProdigy:
yeah, I'd much rather prefer Curve25519 or Ed448-Goldilocks but they require [openssl 1.1.1](https://www.openssl.org/blog/blog/2018/09/11/release111/) (which clients are unlikely to have updated to yet)...

iProdigy:
We should keep an eye on this ticket: https://community.openvpn.net/openvpn/ticket/1116

PR #315, opened 2018-09-28
Add secp256k1 option

iProdigy:
For those of us who would prefer it; https://safecurves.cr.yp.to/rigid.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ OpenVPN 2.4 added support for ECDSA. Elliptic curve cryptography is faster, ligh
 
 This script provides:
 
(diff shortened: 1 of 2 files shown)


iProdigy:
Any downsides of allowing users this choice (at least until better options are [available](https://github.com/angristan/openvpn-install/issues/302#issuecomment-425240686))?

jellemdekker:
I think it would be better if you specified the new curve as option 4, so that the first three curves are the NIST curves, and the fourth is the Koblitz curve. That way the grouping is more logical to the user.

iProdigy:
I'd be happy to if @angristan would be willing to merge it in that order. (I went for this order to keep '1)' as the recommended option and list the rest in increasing size of key bits)

PR #318, opened 2018-10-01
Listen on a specific IPv4

xopez:
Needed if you have more than one IPv4. Just listen on the configured IP on the install.
Tested in my environment.
I have apache server running on one interface with port 443. So its not bindable there. I have bought a second IP. Now OpenVPN needs to know which IP he should bind and not every. So simply setting it in the config. Without this I couldn't connect. Edited config and worked fine for me.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -660,6 +660,7 @@ function installOpenVPN () {
 	chmod 644 /etc/openvpn/crl.pem
 
 	# Generate server.conf
+	echo "local $IP" >> /etc/openvpn/server.conf
 	echo "port $PORT" > /etc/openvpn/server.conf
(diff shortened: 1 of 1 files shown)


angristan:
Thanks for the PR.

This option was added in https://github.com/angristan/openvpn-install/pull/91 and removed by me in https://github.com/angristan/openvpn-install/commit/a7a277e2dc5e6d77a355a4cb60bb85032d724c65.

Search for `Socket bind failed on local address [AF_INET] IP:1194 Cannot assign requested address` and you will a lot of people having the issue.

Since this is a specific use case, maybe adding the TIP in the FAQ would be enough!

xopez:
didn't saw that. Just closing it.

PR #349, opened 2018-11-01
Added Amazon AMI support

BenMenking:
Allows script to run and install necessary components for Amazon Linux

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -67,7 +67,7 @@ function checkOS () {
 			fi
 		fi
 		OS=centos
-	elif [[ -e /etc/os-release ]]; then
+	elif [[ -e /etc/system-release && `grep -is "Amazon Linux" /etc/system-release` ]]; then
 		OS=amazon
 	elif [[ -e /etc/arch-release ]]; then
 		OS=arch
@@ -1147,6 +1147,8 @@ function removeOpenVPN () {
 			yum remove -y openvpn
 		elif [[ "$OS" = 'fedora' ]]; then
 			dnf remove -y openvpn
+		elif [[ "$OS" = 'amazon' ]]; then
+			yum -y remove openvpn
 		fi
 
 		# Cleanup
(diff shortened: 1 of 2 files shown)


openvpn-install.sh
```diff
@@ -572,6 +574,9 @@ function installOpenVPN () {
 		fi
 		# Ubuntu > 16.04 and Debian > 8 have OpenVPN >= 2.4 without the need of a third party repository.
 		apt-get install -y openvpn iptables openssl wget ca-certificates curl
+	elif [[ "$OS" = 'amazon' ]]; then
+		amazon-linux-extras install epel -y
```
bytestream, line 578: `./openvpn-install.sh: line 578: amazon-linux-extras: command not found`

> If you receive an error stating sudo: amazon-linux-extras: command not found, then your instance was not launched with an Amazon Linux 2 AMI (perhaps you are using the Amazon Linux AMI instead). You can view your version of Amazon Linux with the following command. 

````bash
$ cat /etc/system-release
Amazon Linux AMI release 2018.03
````

bytestream: Also `systemd` isn't available on Amazon Linux 1 so some of the later commands generate errors

````
cp: cannot stat ‘/lib/systemd/system/openvpn@.service’: No such file or directory
sed: can't read /etc/systemd/system/openvpn@.service: No such file or directory
sed: can't read /etc/systemd/system/openvpn@.service: No such file or directory
./openvpn-install.sh: line 841: systemctl: command not found
./openvpn-install.sh: line 842: systemctl: command not found
./openvpn-install.sh: line 843: systemctl: command not found
./openvpn-install.sh: line 899: /etc/systemd/system/iptables-openvpn.service: No such file or directory
./openvpn-install.sh: line 902: systemctl: command not found
./openvpn-install.sh: line 903: systemctl: command not found
./openvpn-install.sh: line 904: systemctl: command not found
````

angristan: systemd is required by this script, we can't support AL 1.

PR #368, opened 2018-12-15
Binding server to a specific IP address

MrDanao:
The server may have multiple network interfaces or multiple IP addresses on a same network interface.

On the server configuration, by adding `local $IP` option, it specifies to bind on the IP address `$IP`.

Use case example:
- the server has 2 public IPs
- a web server is listening on the first IP address on TCP/443
- an openvpn server is listening on the second IP address on TCP/443

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -673,7 +673,8 @@ function installOpenVPN () {
 		echo "proto ${PROTOCOL}6" >> /etc/openvpn/server.conf
 	fi
 
-	echo "dev tun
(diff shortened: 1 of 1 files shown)


angristan:
Doesn't work for me:

```
RESOLVE: Cannot resolve host address: <IP>:1194 (Address family for hostname not supported)
```

I remember removing this option: https://github.com/angristan/openvpn-install/commit/a7a277e2dc5e6d77a355a4cb60bb85032d724c65

PR #371, opened 2018-12-19
Add connection to server through IPv6 interface

tuckyapps:
Adds option to establish connection with openvpn server through IPv6. This way, if your ISP is already giving you an IPv6 address, you will still be able to connect to your server.
The IPv6 address is added on client's configuration file, below the IPv4 remote instruction. This way, a user will always try to connect to the IPv4 one, and only if it has no luck, it will then try to connect using IPv6.

Note: When using udp, client will need to wait until keepalive time elapses before it tries to connect through IPv6. If you would like this switch to be fast, I suggest you use TCP protocol instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ This script will let you setup your own secure VPN server in just a few seconds.
 First, get the script and make it executable :
 
 ```bash
-curl -O https://raw.githubusercontent.com/Angristan/openvpn-install/master/openvpn-install.sh
+curl -O https://raw.githubusercontent.com/tuckyapps/openvpn-install/master/openvpn-install.sh
 chmod +x openvpn-install.sh
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -268,6 +267,6 @@ You can [say thanks](https://saythanks.io/to/Angristan) if you want!
 
 ## Credits & Licence
 
-Many thanks to the [contributors](https://github.com/Angristan/OpenVPN-install/graphs/contributors) and Nyr's original work.
+Many thanks to the [contributors](https://github.com/tuckyapps/OpenVPN-install/graphs/contributors) and Nyr's original work.
```
robertkraig, line 270: You should remove all references to `tuckyapps` if you're requesting a P.R. like this. You contribute to the Project, you don't change it to go to your repo.

robertkraig (changes requested):

README.md
ghost: Agreed.

ThePigPeppa:
this is a bad idea.

tuckyapps:
Hi!

What do you mean?

What should I do?

Thanks in advance,

Best,

Santiago Rosenblatt

angristan:
Hello,

Can you rebase your PR against master?

tuckyapps:
How about now?

I'm sorry, did not know my commits were still influencing the pull request.

Back in January, I was asked by my company to update the comments so they
point to my repo in case others would like to download it, increasing
security and more control over the code.

Is it okay now?

Thanks again!

angristan:
Hello,

the diff is really messy right now, I would advise starting again from master, checking out a new branch, then adding the modifications regarding this PR then force-pushing the branch, please 🙏

PR #384, opened 2019-01-21
Update README.md

ghost:


randshell:
Usually you look at the script before running it and not pipe it to bash like this.

A lot of sites have this warning around, for example the docker install script at https://docs.docker.com/install/linux/docker-ce/debian/#install-using-the-convenience-script

Note also that to add or revoke clients, uninstall openvpn etc. you need to run the script again, not possible if it's not downloaded.

angristan:
Agreed.

Issue #261, opened 2018-07-18
Fully automatic installation (without user input)


sayem314:
Update codes to support advanced/automatic installation. This step will reduce the time of installation for the users that install very often. This will suppress any prompts as part of automatic installation process.

Here is an example how it will be done:

    auto=yes ./openvpn-install.sh

And all the default values will be used for installation process. Also I will add the ability to add custom option too!

sayem314:
@Angristan are you interested to see this?

angristan:
Yes I'm interested, although I've considered this out-of-scope until now... 

If this is added to the script it has to be clean, maintainable and not add bloat.

This is not my priority, but if a PR is made I'll consider adding it to the script.

sayem314:
I lost my interest over the month.

/s

cliffcotino:
I created a pull request to be able to support this #390

angristan:
I merged the PR made by @cliffcotino. It was easier than expected :)

angristan:
I think it would be even better to be able to do:

```sh
export AUTO_INSTALL=y
./openvpn-install.sh
```

And then a simple condition a the beginning of the script would set all these variables for us (in place of the manual exports)

angristan:
Note that password-protected clients are not supported by this, because of easy-rsa. I have not found a workaround for this.

luizluca:
easyrsa can now accept passwords as arguments:
https://github.com/OpenVPN/easy-rsa/commit/c794d5ebaae7ec911e638a4f2c21137a3fb5ec9c

Issue #319, opened 2018-10-02
Auto update script


chocolateshirt:
It would be great if there is feature to get the latest update of the script without download latest script and reinstall it manually.

The example, if you run `./openvpn-install.sh` after installation, there will be new menu like this:

```
Welcome to OpenVPN-Install!

OpenVPN-Install ver. xxx
The git repository is available at: https://github.com/angristan/openvpn-install

What do you want to do?
   1) Add a new user
   2) Revoke existing user
   3) Update script
   4) Remove OpenVPN
   5) Exit
Select an option [1-5]:
```

FevLoad:
I second this opinion, that would be great if the coder could consider doing it.

xiagw:
Yes, have a good reference: [https://github.com/Neilpang/acme.sh](https://github.com/Neilpang/acme.sh)
Can update itself.

xiagw:
[git@github.com:Neilpang/acme.sh.git](url)

Leopere:
At the moment all you need to do currently is do a guy pull.

I mean I guess it would be trivial to add something that runs that command
for you but it seems moot-ish

xiagw:
Ref: [https://github.com/alexanderepstein/bsupdate](https://github.com/alexanderepstein/bsupdate)

Firerouge:
Updating the script doesn't help unless it can also upgrade the generated configs made by older versions of the script as well.

That'd largely be changing existing lines in the ovpn files, and prompting the user if they'd like to enable newly available options. Which means you have to push your new ovpn configs to any and all clients to see most of the benefits.

I think there's something to be said for regenerating the configuration files while persisting your current keys. But simply updating the script (as in a git pull) as described here would still require you to remove openvpn and rerun the install script to get the benefits of changes to the script

Leopere:
What if there was just an answers.env file that would be saved with the responses given to the installer and then the update script would just uninstall and reinstall with the answers and potential backups of certs etc.

angristan:
See https://github.com/angristan/openvpn-install/issues/179#issuecomment-389820907

randshell:
We all commented about some upgrade code but after reading Pivpn's [wiki](https://github.com/pivpn/pivpn/wiki/Upgrading) I understand @angristan has a point, even if he didn't go in as much details.

That's about the code. Regarding the process of upgrading, even if manual, I think it _could_ be a little more friendly for the user if some more work is done from the dev side.
[...]

PR #390, opened 2019-02-02
Headless installation

cliffcotino:
I've made some small changes to be able to support automatic installation (#261)
It does not support adding a password protected client, or installation behind a NAT, but it's a start.

For installation behind a NAT, the IP detection like could be changed to something like:
`IP=$(wget -q -O - https://api.ipify.org)`

Auto installation command:
```
export APPROVE_INSTALL=y
export APPROVE_IP=y
export IPV6_SUPPORT=n
export PORT_CHOICE=2
export PORT=11194
export PROTOCOL_CHOICE=1
export DNS=9
export COMPRESSION_ENABLED=n
export CUSTOMIZE_ENC=n
export CLIENT=clientName
export PASS=1
./openvpn-install.sh
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ If you have any question, head to the [FAQ](#faq) first.
 
 ### Headless install
 
-It's also possible to run the script headless.
+It's also possible to run the script headless, e.g. without waiting for user input, in an automated manner.
 
 Example usage:
 ```bash
@@ -51,10 +51,12 @@ export PROTOCOL_CHOICE=1
 export DNS=1
 export COMPRESSION_ENABLED=n
 export CUSTOMIZE_ENC=n
-export CLIENT=clientName
+export CLIENT=clientname
 export PASS=1
 ./openvpn-install.sh
 ```
+
+If the server is behind NAT, you can specify its endpoint with the `PUBLICIP` variable. It the endpoint is the public IP address which it is behind, you can use `export PUBLICIP=$(curl ifconfig.co)`.
 
 ## Features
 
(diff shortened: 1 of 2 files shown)


randshell:
Hi! Thank you for the headless install PR. Indeed it allows for easier reinstall, for example when this script is updated, because it allows to save the variables needed ( we should say it in the README too and maybe ask the user on the first run to export them in a file ) and do its work.

> For installation behind a NAT, the IP detection like could be changed to something like:
 `IP=$(wget -q -O - https://api.ipify.org)`

It should be `$PUBLICIP` instead and so can be exported by the user.
But I don't think using ipify it's a good idea, behind NAT are home users that also have a dynamic IP. In this case the solution is to export the DDNS domain and not use a temporary dynamic IP address to generate the configs.

cliffcotino:
I'll update the README with some instructions how to run the script using a headless install.
I'll see what I can come up with for the option to export it on the first run.

You're right about the `$PUBLICIP` variable, I was thinking about my own use case a bit too much 😃

randshell:
Exporting the variables on first run for easier reinstall was only an idea that for @angristan may be out-of-scope. I was thinking at https://github.com/angristan/openvpn-install/issues/319#issuecomment-449677275 and the related issue. So he decides when he has some free time.

cliffcotino:
@angristan I pushed some documentation on how to use the headless install.
Please let me know if there are other changes you would like to see in this PR

angristan:
Thank you, @cliffcotino.

PR #397, opened 2019-02-26
Fix Variable Substitution for ENDPOINT

LopezNathan:
This corrects an issue with the ENDPOINT variable utilizing the word "PUBLIC_IPV4" instead of the value from variable `PUBLIC_IPV4`.


angristan (approved):

angristan:
Thank you

Issue #356, opened 2018-11-23
Add support for firewalld (CentOS-based distributions)


angristan:
From https://github.com/angristan/openvpn-install/issues/348#issuecomment-438724343

koallen:
I also had this problem. It seems that masquerade was not enabled at all after running the install script. The ONLY thing I did was to manually add masquerade with `firewall-cmd` and the VPN has Internet access.

angristan:
Thanks for the feedback.

BornToBeRoot:
> I also had this problem. It seems that masquerade was not enabled at all after running the install script. The ONLY thing I did was to manually add masquerade with `firewall-cmd` and the VPN has Internet access.

This worked for me too. But `firewall-cmd [--permanent] --add-masquerade` gives me strange issues on openvz... i can enable it, but not disable. I have to use `firewall-offline-cmd` to reverse the changes.

I tried the script from [Nyr](https://github.com/Nyr/openvpn-install) and there are no problems with CentOS 7 on openvz. The Internet connection works, even when masquerade is not enabled via `firewall-cmd`.

@angristan would be great if you an take a look. I would like to use your script because of its security features.

BornToBeRoot:
https://github.com/Nyr/openvpn-install/blob/d4efae3b1081867ffce420267f8c20a7c7336047/openvpn-install.sh#L318

```
if pgrep firewalld; then
firewall-cmd --zone=public --add-port=$PORT/$PROTOCOL
firewall-cmd --zone=trusted --add-source=10.8.0.0/24
firewall-cmd --permanent --zone=public --add-port=$PORT/$PROTOCOL
firewall-cmd --permanent --zone=trusted --add-source=10.8.0.0/24
# Set NAT for the VPN subnet
firewall-cmd --direct --add-rule ipv4 nat POSTROUTING 0 -s 10.8.0.0/24 ! -d 10.8.0.0/24 -j SNAT --to $IP
firewall-cmd --permanent --direct --add-rule ipv4 nat POSTROUTING 0 -s 10.8.0.0/24 ! -d 10.8.0.0/24 -j SNAT --to $IP
```

angristan:
PR to merge -> https://github.com/angristan/openvpn-install/pull/417

PR #417, opened 2019-04-29
Detect firewalld and use firewall-cmd (fix for #356)

BornToBeRoot:
If firewalld is used (eg on CentOS 7), the firewall configuration is not correct (nat, internet access)... #356 

When firewalld is running on the system, the following rules will be applied:

```
if [[ "$PORT" == '1194' ]]; then
    firewall-cmd --zone=public --remove-service=openvpn
    firewall-cmd --permanent --zone=public --remove-service=openvpn
else
    firewall-cmd --zone=public --remove-port=$PORT/$PROTOCOL
    firewall-cmd --permanent --zone=public --remove-port=$PORT/$PROTOCOL
fi          

firewall-cmd --zone=trusted --remove-source=10.8.0.0/24
firewall-cmd --permanent --zone=trusted --remove-source=10.8.0.0/24

firewall-cmd --direct --remove-rule ipv4 nat POSTROUTING 0 -s 10.8.0.0/24 ! -d 10.8.0.0/24 -j SNAT --to $IP
firewall-cmd --permanent --direct --remove-rule ipv4 nat POSTROUTING 0 -s 10.8.0.0/24 ! -d 10.8.0.0/24 -j SNAT --to $IP
```

In the removed function, they are deleted:

```
# Allow incoming traffic
if [[ "$PORT" == '1194' ]]; then
    firewall-cmd --zone=public --add-service=openvpn
    firewall-cmd --permanent --zone=public --add-service=openvpn
else
    firewall-cmd --zone=public --add-port=$PORT/$PROTOCOL
    firewall-cmd --permanent --zone=public --add-port=$PORT/$PROTOCOL
fi      

 # Add trusted zone
firewall-cmd --zone=trusted --add-source=10.8.0.0/24		
firewall-cmd --permanent --zone=trusted --add-source=10.8.0.0/24

# Set NAT for the VPN subnet
firewall-cmd --direct --add-rule ipv4 nat POSTROUTING 0 -s 10.8.0.0/24 ! -d 10.8.0.0/24 -j SNAT --to $IP
firewall-cmd --permanent --direct --add-rule ipv4 nat POSTROUTING 0 -s 10.8.0.0/24 ! -d 10.8.0.0/24 -j SNAT --to $IP
```

I copied these rules from Nyr/openvpn-install.

The rules with firewall-cmd are applied permanent and not with the systemd script/service... I see some problems if you do this with the script:
1) If you don't use the parameter `--permanent` in the script, the rules are only temporary. When you restard the firewalld service or use `firewall-cmd --reload`, the configuration is lost until you restart `iptables-openvpn`
2) If you use `--permanent` in the script and openvpn is uninstalled with this script, the rules are still active until you restart firewalld or reload the configration.

Maybe you can add both, the temporary and the permanent rule in the service. For my needs (an always running vpn), i don't see the added value of this systemd service.

I tested the script only on my CentOS 7.


openvpn-install.sh
```diff
@@ -852,44 +852,63 @@ verb 3" >> /etc/openvpn/server.conf
...
+    # Add firewall rules --> firewalld / iptable (systemd scripts)
+    if pgrep firewalld; then
+        # Allow incoming traffic
+		if [[ "$PORT" == '1194' ]] && [[ "$PROTOCOL" == "udp" ]]; then
+			firewall-cmd --zone=public --add-service=openvpn
+            firewall-cmd --permanent --zone=public --add-service=openvpn
+        else
+            firewall-cmd --zone=public --add-port="$PORT/$PROTOCOL"
+            firewall-cmd --permanent --zone=public --add-port="$PORT/$PROTOCOL"
+		fi
```
randshell, line 864:
```suggestion
            # Define new openvpn service
            cp /usr/lib/firewalld/services/openvpn.xml /etc/firewalld/services
            sed -i "s|udp|${PROTOCOL}|" /etc/firewalld/services/openvpn.xml
            sed -i "s|1194|${PORT}|" /etc/firewalld/services/openvpn.xml
                    
            firewall-cmd --zone=public --add-service=openvpn
            firewall-cmd --permanent --zone=public --add-service=openvpn
```

openvpn-install.sh
```diff
@@ -1124,14 +1144,37 @@ function removeOpenVPN () {
...
+			# Get OpenVPN protocol from the configuration
+			PROTOCOL=$(grep '^proto ' /etc/openvpn/server.conf | cut -d " " -f 2)
+
+            if [[ "$PORT" == '1194' ]] && [[ "$PROTOCOL" == "udp" ]]; then
+                firewall-cmd --zone=public --remove-service=openvpn
+                firewall-cmd --permanent --zone=public --remove-service=openvpn
+            else
+                firewall-cmd --zone=public --remove-port="$PORT/$PROTOCOL"
+                firewall-cmd --permanent --zone=public --remove-port="$PORT/$PROTOCOL"
+		    fi
```
randshell, line 1160:
```suggestion
            firewall-cmd --zone=public --remove-service=openvpn
            firewall-cmd --permanent --zone=public --remove-service=openvpn
            
            # Remove defined openvpn service
            rm /etc/firewalld/services/openvpn.xml
```

randshell (changes requested):
I suggest defining a new openvpn service instead of adding port and protocols without a name or description.

Changes are not tested.

openvpn-install.sh
```diff
@@ -852,44 +852,63 @@ verb 3" >> /etc/openvpn/server.conf
...
-	echo "#!/bin/sh
+    # Add firewall rules --> firewalld / iptable (systemd scripts)
+    if pgrep firewalld; then
+        # Allow incoming traffic
+		if [[ "$PORT" == '1194' ]] && [[ "$PROTOCOL" == "udp" ]]; then
+			firewall-cmd --zone=public --add-service=openvpn
+            firewall-cmd --permanent --zone=public --add-service=openvpn
+        else
+            firewall-cmd --zone=public --add-port="$PORT/$PROTOCOL"
+            firewall-cmd --permanent --zone=public --add-port="$PORT/$PROTOCOL"
```
angristan, line 863: Just keep the else.

```suggestion
            firewall-cmd --zone=public --add-port="$PORT/$PROTOCOL"
            firewall-cmd --permanent --zone=public --add-port="$PORT/$PROTOCOL"
```

openvpn-install.sh
```diff
@@ -1124,14 +1144,37 @@ function removeOpenVPN () {
...
+			IP=$(firewall-cmd --direct --get-rules ipv4 nat POSTROUTING | grep '\-s 10.8.0.0/24 '"'"'!'"'"' -d 10.8.0.0/24 -j SNAT --to ' | cut -d " " -f 10)
+			# Get OpenVPN protocol from the configuration
+			PROTOCOL=$(grep '^proto ' /etc/openvpn/server.conf | cut -d " " -f 2)
+
+            if [[ "$PORT" == '1194' ]] && [[ "$PROTOCOL" == "udp" ]]; then
+                firewall-cmd --zone=public --remove-service=openvpn
+                firewall-cmd --permanent --zone=public --remove-service=openvpn
+            else
+                firewall-cmd --zone=public --remove-port="$PORT/$PROTOCOL"
+                firewall-cmd --permanent --zone=public --remove-port="$PORT/$PROTOCOL"
```
angristan, line 1159: Same here
```suggestion
                firewall-cmd --zone=public --remove-port="$PORT/$PROTOCOL"
                firewall-cmd --permanent --zone=public --remove-port="$PORT/$PROTOCOL"
```

angristan (changes requested):

angristan:
Thank you for this!

randshell:
@angristan what about my suggestion of defining a new service to keep things clean? You resolved it without an explanation.

Thanks

randshell:
@angristan firewalld was supported but you removed it in #291. Why?

angristan:
@randomshell sorry, that seems like a cleaner solution indeed

https://github.com/angristan/openvpn-install/pull/291 was about handling all the firewall rules with systemd and iptables. I dropped firewalld for the sake of simplicity, and as written in a comment I'm pretty sure I tested these modifications on all distributions, even after reboots. Not sure what changed since then.

randshell:
As a note, with SNAT if the IP changes then users can't connect anymore. Should we use MASQUERADE in firewalld like we do with iptables or update the iptables commands instead?

Source https://unix.stackexchange.com/questions/21967/difference-between-snat-and-masquerade

randshell:
@BornToBeRoot could you please review my initial suggestions of the new service and update your PR?

This issue is stale for a year and blocks our support for CentOS

BornToBeRoot:
@randomshell i updated the PR.

@angristan there are some conflicts that needs to be resolved. Maybe you can take a look at it and then merge the PR.

randshell:
@BornToBeRoot Great!

The only thing missing is a `firewall-cmd --reload` after adding or removing the rules. I know you accepted the suggested changes but I realized that we need a reload after.

jcastilloalonso:
This might bring issues if the setup is done behind a NAT. Ideally we would use Masquerating instead of SNAT to avoid this issue and not mess with all the headache.

PR #433, opened 2019-06-13
formatting code and fix shellcheck error

xiagw:
1. Formatting code.
2. Fix shellcheck error


ghost (commented):
seems pretty trivial and pointless. prior formatting works just fine without running it through shellcheck.

randshell:
> seems pretty trivial and pointless. prior formatting works just fine without running it through shellcheck.

It's not about working. It's about easier reading and maintaining.

Anyway I think we should force a formatting because contributors tend to change it with new PRs @angristan.

randshell (commented):
This PR is outdated and with conflicts, but if it gets considered again I suggest we escape variables too.

PR #448, opened 2019-07-14
Update README: HMAC-SHA1 is not broken

davebarda:
Remove irrelevant data from readme, as HMAC strength does not depend on the collisions of the hash function

For more info:
http://cseweb.ucsd.edu/~mihir/papers/hmac-new.html


angristan:
Thanks! Other links I found:

- https://www.schneier.com/blog/archives/2005/02/sha1_broken.html
- https://crypto.stackexchange.com/questions/26510/why-is-hmac-sha1-still-considered-secure
- https://security.stackexchange.com/questions/33123/hotp-with-as-hmac-hashing-algoritme-a-hash-from-the-sha-2-family
- https://security.stackexchange.com/questions/187866/why-aren-t-collisions-important-with-hmac

PR #449, opened 2019-07-15
Asks for pass for new CA authority

ghost:
Adds extra security in case server is ever breached. If an attacker is able to take your CA private key it is encrypted with a passphrase.

The regular script allows anyone who is able to steal your CA private key to issue infinite new clients for your VPN without any barriers.

Obviously your server should be protected enough your private key is not stolen in the first place, but this adds an extra level of security.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,29 @@
 
 OpenVPN installer for Debian, Ubuntu, Fedora, CentOS and Arch Linux.
 
-This script will let you setup your own secure VPN server in just a few seconds.
+This script will let you setup your own secure VPN server in just a few minutes.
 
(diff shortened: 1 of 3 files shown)


angristan (changes requested):
I'm not sure I like this but at least:

- Don't rename the file (?)
- Make this optional

ghost:
Sorry, renamed the file for my own personal use. I don't expect this PR to be accepted as-is. I do however think it is a good basis for discussion when the aim of this project to create a hardened OpenVPN install.

I agree adding some code to make it optional would be best. And explain to the user the consequences too (the only negative I can tell is having to type the password in when rebooting the server).

PR #456, opened 2019-08-12
Create openvpndialog.sh

techroy23:
Initial Release
retained original functions except manageusers
main function is not yet link


techroy23:
i will try to complete the code tommorow

techroy23:
added first 4 steps of install.
sorry im too slow

techroy23:
will try to complete the $installQuestions tommorow.

angristan (changes requested):
Thanks for your interest, but duplicating the code into another script is completely out of the question. 🙁

Also, that's why I wasn't sure about this, it adds a lot of code the script, which is already very heavy.

PR #426, opened 2019-06-05
Add Amazon Linux 2 support

chafouin:
Add Amazon Linux 2 support.

PR #459, opened 2019-08-19
Add Amazon Linux 2 support

angristan:
From #426


angristan:
It works well 👍

angristan:
Thanks @chafouin

PR #480, opened 2019-09-09
Add Gentoo and OpenRC Support

r7l:
Hello, i've added support for Gentoo Linux. Would it be possible to merge it as this script is really helpful but lacks of support for Gentoo Linux users. Since Gentoo is flexible about the init system, i've also added support for OpenRC. The overhead for the support of 2 init systems isn't too big as i've wrapped a simple function around it.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -1018,18 +1018,21 @@ INSTANCE_NAME="Iptables for OpenVPN"
 
 description="IPTables for OpenVPN"
 
+depends() {
+    need openvpn
+}
+
 start() {
-    ebegin "Starting $INSTANCE_NAME"
+    ebegin "Starting ${SVCNAME}"
     /etc/iptables/add-openvpn-rules.sh
     eend $?
 }
 
 stop() {
-    ebegin "Stopping $INSTANCE_NAME"
+    ebegin "Stopping ${SVCNAME}"
     /etc/iptables/rm-openvpn-rules.sh
     eend $?
 }
-
 ENDOFFILE
 
 chmod +x /etc/init.d/iptables-openvpn
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -1,10 +1,16 @@
 #!/bin/bash
 
-# Secure OpenVPN server installer for Debian, Ubuntu, CentOS, Amazon Linux 2, Fedora and Arch Linux
+# Secure OpenVPN server installer for Debian, Ubuntu, CentOS, Amazon Linux 2, Fedora, Arch Linux and Gentoo Linux
 # https://github.com/angristan/openvpn-install
 
 function isRoot () {
 	if [ "$EUID" -ne 0 ]; then
+		return 1
+	fi
+}
+
+function isSystemD () {
+	if [ ! -e /bin/systemctl ]; then
 		return 1
 	fi
 }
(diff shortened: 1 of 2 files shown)


openvpn-install.sh
```diff
@@ -1185,6 +1297,8 @@ function removeOpenVPN () {
 			fi
 		elif [[ "$OS" = 'arch' ]]; then
 			pacman --noconfirm -R openvpn
+		elif [[ "$OS" = 'arch' ]]; then
```
angristan: Should be `gentoo` here I guess?

angristan:
I wouldn't usually agree to that, but first of all, thank you for doing things properly! 👍

r7l:
Thanks for your response. I've fixed that issue with the uninstall. Didn't really notice as the scripts installs OpenVPN anyways each time you redo the process.

I've also added Gentoo into this description for the error and updated Readme.md. Somehow i forgot about this.

Surely it is your project but the support for Gentoo and OpenRC doesn't seem like a big change to the script for me. It's just the additional init function, few replaced strings and a couple more IFs.

r7l:
Is there any chance of this ever being merged or you decided against it?

angristan:
Sorry, I'm abroad so I have less time for GitHub stuff as you can see on my graph 🙂

I also I have to see how to setup a Gentoo VM to test this PR...

angristan:
Gentoo is not supported natively on:

- Digital Ocean
- Vultr
- GCP
- AWS
- Hetzner
- ...

That's why I was against this initially. First, that means very few people would actually use the script on Gentoo. Secondly, that means I can't easily test it on Gentoo, and that's a problem...

tidux:
AWS AMIs: http://www.dowdandassociates.com/products/pygoscelis-papua/

angristan:
Thanks @tidux

angristan:
Errors I see:

```
+ sed -i 's/^RANDFILE/#RANDFILE/g' pki/openssl-easyrsa.cnf
sed: can't read pki/openssl-easyrsa.cnf: No such file or directory
```

```
+ /etc/init.d/iptables-openvpn start
 * Caching service dependencies ...                                                                                                                                         [ ok ]
 * Starting  ...                                                                                                                                                            [ !! ]
 * ERROR: iptables-openvpn failed to start
```

openvpn-install.sh
```diff
@@ -938,10 +1009,40 @@ RemainAfterExit=yes
...
+description="IPTables for OpenVPN"
+
+depends() {
+    need openvpn
+}
+
+start() {
+    ebegin "Starting ${SVCNAME}"
+    /etc/iptables/add-openvpn-rules.sh
+    eend $?
```
angristan, line 1028: This results in `eend 1` being written, so the init script fails

openvpn-install.sh
```diff
@@ -938,10 +1009,40 @@ RemainAfterExit=yes
...
+}
+
+stop() {
+    ebegin "Stopping ${SVCNAME}"
+    /etc/iptables/rm-openvpn-rules.sh
+    eend $?
```
angristan, line 1034: Same here

angristan (changes requested):

angristan:
This workaround should be skipped on Gentoo since it does not seem to be affected:

```sh
# Workaround to remove unharmful error until easy-rsa 3.0.7
# https://github.com/OpenVPN/easy-rsa/issues/261
sed -i 's/^RANDFILE/#RANDFILE/g' pki/openssl-easyrsa.cnf
```

PR #489, opened 2019-09-18
Making Client Revocation Headless and Documenting New Feature

### LINKED ISSUE #486: Make Client Revocation Headless
Currently, we have a headless task for
 - installing the entire OpenVPN server as well as 
 - adding a single user
The only thing that's missing right now is the _headless revocation of client certificates_.

This should be added as it allows to automate another task, especially as reusing existing coding techniques should make this fairly straightforward.

And while we're at it, we should also document the new feature. Because documentation matters.

dumrauf:
This closes https://github.com/angristan/openvpn-install/issues/486

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,19 @@ Other variables can be set depending on your choice (encryption, compression). Y
 
 Password-protected clients are not supported by the headless installation method since user input is expected by Easy-RSA.
 
+### Headless User Removal
+
+The removal of an existing user can also be fully automated. Again, the key is to provide the (string) value of the `MENU_OPTION` variable along with the remaining mandatory variables before invoking the script.
+
+The following Bash script removes the existing user `bar` from an OpenVPN configuration
+```bash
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -83,6 +83,18 @@ export PASS="1"
 ./openvpn-install.sh
 ```
 
+### Headless User Removal
+
+The removal of an existing user can also be fully automated. Again, the key is to provide the (string) value of the `MENU_OPTION` variable along with the remaining mandatory variables before invoking the script.
+
+The following Bash script removes the existing user `bar` from an OpenVPN configuration
+```bash
+#!/bin/bash
```
angristan, line 92: That shouldn't be here

angristan (changes requested):

README.md
randshell: Why? It's the same as the example above.

angristan:
Thanks, this is supported now

PR #492, opened 2019-09-22
Update supported architectures

rhysperry111:
I have got it working on my raspi (ArchlinuxArm - armv6l) without any trouble or additional steps. Thanks for such a great tool!


angristan:
Thanks!

Issue #495, opened 2019-09-24
Fedora 30 requires change


pswilde:
Configuring on Fedora Server 30, haven't tried earlier versions, all works fine once configured but a minor change is needed in the script.
Because Fedora Server 30 has both /etc/system-release and /etc/fedora-release files, and /etc/system-release gets checked first in the script, the $OS variable in checkOS() returns an empty value. Moving the `elif [[ -e /etc/fedora-release ]]; then
                OS=fedora` line before the `elif [[ -e /etc/system-release ]]; then` line resolves the issue for now.

angristan:
Seems to have been fixed in the meantime

PR #501, opened 2019-10-12
Test for fedora earlier

rrobgill:
As reported in Bug #495 , fedora 30+ includes both /etc/os-release and  /etc/fedora-release. Current order tests os-release first, and does not correctly identify fedora 30+ installations.

This change moves the check for /etc/fedora-release prior to those for /etc/os-release.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -50,6 +50,8 @@ function checkOS () {
 				fi
 			fi
 		fi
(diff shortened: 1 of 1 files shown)


angristan:
That seems OK for me, it doesn't affect CentOS or Amazon Linux right?

angristan:
Seems to have been fixed in the meantime

PR #507, opened 2019-10-20
Added openvpn.py

ivica3730k:
Added ability to setup OVPN using Python script. 
Also added ability to use google gmail account to send .OVPN file to end user.
(easier and more secure that to move it to webserver for download)
Ovpn profile will be created based on email account, profile name would be prefix of the email address.

Also updated .gitignore to ignore .ovpn files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 .vagrant/
 *.log
+.ovpn
+.bash
--- a/openvpn.py
+++ b/openvpn.py
@@ -1,0 +1,96 @@
+import sys
+import os
+import subprocess
+import smtplib
+from os.path import basename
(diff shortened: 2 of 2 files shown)


TerrorBite:
I only have one thing to say:

    python openvpn.py --email "-rf /"

LouisNitag:
More fun with this :laughing: 

```python openvpn.py --email "; echo newpassword | passwd --stdin root;"```

ghost:
Lol why close the pr vs update with some proper sanitization?

angristan:
Thank you for your work. This can be useful as a separate wrapper script, but it does not fit the `openvpn-install` project.

PR #550, opened 2020-02-28
Custom IP Range

browningluke:
I wanted to get the server OpenVPN to run on an IP range that was not `10.8.0.0`. I added the ability for the user to specify which range the server shall assign IP addresses. This does not, however, implement custom subnets, it uses the default /24.

My changes are below. You can find these at line 246 in the file `openvpn-install.sh`. I followed a similar variable name as the others. This variable is then used in place of `10.8.0.0` and `10.8.0.1`.

```bash
echo "What IP range settings do you want OpenVPN to use?"
echo "   1) Default: 10.8.0.0"
echo "   2) Custom"
until [[ "$IP_CHOICE" =~ ^[1-2]$ ]]; do
	read -rp "IP choice [1-2]: " -e -i 1 IP_CHOICE
done
case $IP_CHOICE in
	1)
		IP_RANGE="10.8.0.0"
	;;
	2)
		until [[ "$IP_RANGE" =~ ^([0-9]{1,3}\.){3}0$ ]]; do
			read -rp "Custom IP [x.x.x.0]: " -e -i 10.8.0.0 IP_RANGE
		done
	;;
esac
```


brnl:
I don't understand the variable $VAR. What's it supposed to mean? Isn't there a better name?

browningluke:
Yes it is supposed to have a better name, I sincerely apologize for that error. I committed the wrong file, `${VAR: : -1}` is supposed to be `${IP_RANGE: : -1}`. I have fixed this is my latest commit.

Let me know if you have any other concerns.

openvpn-install.sh
```diff
@@ -911,7 +929,7 @@ ip6tables -I FORWARD 1 -i tun0 -o $NIC -j ACCEPT" >> /etc/iptables/add-openvpn-r
 
 	# Script to remove rules
 	echo "#!/bin/sh
-iptables -t nat -D POSTROUTING -s 10.8.0.0/24 -o $NIC -j MASQUERADE
+iptables -t nat -D POSTROUTING -s #IP_RANGE/24 -o $NIC -j MASQUERADE
```
randomshell:
```suggestion
iptables -t nat -D POSTROUTING -s $IP_RANGE/24 -o $NIC -j MASQUERADE
```

openvpn-install.sh
```diff
@@ -723,7 +741,7 @@ persist-key
 persist-tun
 keepalive 10 120
 topology subnet
-server 10.8.0.0 255.255.255.0
+server $IP_RANGE 255.255.255.0
```
randomshell:
```suggestion
server ${IP_RANGE} 255.255.255.0
```

openvpn-install.sh
```diff
@@ -896,7 +914,7 @@ verb 3" >> /etc/openvpn/server.conf
 
 	# Script to add rules
 	echo "#!/bin/sh
-iptables -t nat -I POSTROUTING 1 -s 10.8.0.0/24 -o $NIC -j MASQUERADE
+iptables -t nat -I POSTROUTING 1 -s $IP_RANGE/24 -o $NIC -j MASQUERADE
```
randomshell:
```suggestion
iptables -t nat -I POSTROUTING 1 -s ${IP_RANGE}/24 -o $NIC -j MASQUERADE
```

openvpn-install.sh
```suggestion
server ${IP_RANGE} 255.255.255.0
```

openvpn-install.sh
randshell, line 986:
```suggestion
iptables -t nat -I POSTROUTING 1 -s ${IP_RANGE}/24 -o ${NIC} -j MASQUERADE
```

openvpn-install.sh
```suggestion
iptables -t nat -D POSTROUTING -s ${IP_RANGE}/24 -o ${NIC} -j MASQUERADE
```
`#` should be `$`

randshell (changes requested):
The escaping is for best practice.
The requested change is the logic error in line 932.

browningluke:
I have committed your suggested changes. Thank you for the fix.

randshell (approved):
Thank you for contributing!

HenryNe (approved):
Nice. Tesed on Debian 9.

dnk8n:
This would be handy. Any docs on how to change already set tun0 10.8.0.1?

E.g.

```
tun0: flags=4305<UP,POINTOPOINT,RUNNING,NOARP,MULTICAST>  mtu 1500
        inet 10.8.0.1  netmask 255.255.255.0  destination 10.8.0.1
        inet6 xxx::xxx:xxx:xxx:xxx  prefixlen 64  scopeid 0x20<link>
        unspec 00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00  txqueuelen 100      
```

dnk8n:
Also, worth maybe changing the default. See here, https://community.openvpn.net/openvpn/wiki/AvoidRoutingConflicts

The default has a high chance of conflicting with `"default settings for routers/acces points/switches/etc"`

browningluke:
> 
> 
> This would be handy. Any docs on how to change already set tun0 10.8.0.1?
> 
> E.g.
> 
> ```
> tun0: flags=4305<UP,POINTOPOINT,RUNNING,NOARP,MULTICAST>  mtu 1500
>         inet 10.8.0.1  netmask 255.255.255.0  destination 10.8.0.1
>         inet6 xxx::xxx:xxx:xxx:xxx  prefixlen 64  scopeid 0x20<link>
>         unspec 00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00  txqueuelen 100      
> ```

@dnk8n I'm not 100% sure how to accomplish this, I'll have a look into changing an already setup tun0.

> 
> 
> Also, worth maybe changing the default. See here, https://community.openvpn.net/openvpn/wiki/AvoidRoutingConflicts
> 
> The default has a high chance of conflicting with `"default settings for routers/acces points/switches/etc"`

I wouldn't be opposed to changing the default; its quite easy in the code. I only left it as 10.8.0.0 as that is what is currently is in the script.

ckcr4lyf:
This works perfectly, I would really like to see it merged into the main script. Just to dive into my usecase:

I have two VPSs in two separate countries, I wanted to setup OpenVPN on each of them, and then have my local pfSense instance be a client to both of them, t ohave two LANs behind each VPN. For the routing in pfsense, if both clients are on the same `10.8.0.0/24` subnet, it won't work. With this I can easily have my 2nd OpenVPN instance be a different subnet and make it work.

P-a-d-r-a-i-g:
Yes, would like to get this merged into main also. I have multiple VPN servers in different countries and they require simultaneous connections sometimes. Having an option to specify a custom IP instead of 10.8.x.x is ideal

ckcr4lyf (approved):
LGTM

angristan:
Hi! Thanks for the work and thanks for the wait :) This is available since https://github.com/angristan/openvpn-install/commit/e9deb4b8ab2bf1e7138da26fb2306010a904e853

PR #554, opened 2020-03-02
Install semanage command on CentOS

### LINKED ISSUE #553: installing openvpn server on another port
when I installed OpenVPN server with default port 1194 it works properly
but when I want to install it on another port the connection does not connect
I think it is something about open ports or... but I have not any knowledge to solve it
what should I do?

also if I want users to have user and password, what should I do?

I am using centos 7 and latest version of the script, also I updated all modules in Linux

randshell:
CentOS has selinux enabled by default but it hasn't the `semanage` command required to run OpenVPN on another port.

`policycoreutils-python*` match `policycoreutils-python` in CentOS 7 and `policycoreutils-python-utils` in Centos 8.

Fix #553.


angristan:
Thanks!

PR #559, opened 2020-03-10
Custom DNS option wrong value fix

xPakrikx:
Custom DNS option wrong value fix


angristan:
Thanks. I forget this one every single time

PR #578, opened 2020-03-26
Get public interface from IPv6

HenryNe:
In a IPv6 only environment, the variable $NIC would be empty and iptables in add-openvpn-rules.sh will fail by missing argument.


angristan:
Thanks!

PR #579, opened 2020-03-26
Add explicit-exit-notify for UDP

HenryNe:
For faster reconnects with UDP is better to send the the explicit-exit-notify to server. With this the server can directly see, that the client will exit.


angristan:
Awesome, thanks!

PR #580, opened 2020-03-26
Fix error message mkdir /etc/iptables

HenryNe:
Fix this error message:
mkdir: cannot create directory ‘/etc/iptables’: File exists


angristan:
Oops! Thanks for the fix!

PR #581, opened 2020-03-26
Print warning about empty public interface

HenryNe:
Warning, if cannot detect public interface, and give user a choice to continue or abord.


angristan:
Very nice. Thanks

Issue #583, opened 2020-03-27
Problem running on OpenVZ #129 again


HenryNe:
The issure #129 again.
```
daemon() failed or unsupported: Resource temporarily unavailable (errno=11)
```
The modifired file `/etc/systemd/system/openvpn@.service` seams not used by Debian 9.
I need to fix the file `/lib/systemd/system/openvpn@.service`, and comment out the **LimitNPROC**, to get it working.

```
# lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 9.9 (stretch)
Release:        9.9
Codename:       stretch
```

HenryNe:
On Debian 9 the copy of unit file `/etc/systemd/system/openvpn@.service` has no effect. You can test the "no effect" by changing the Description in the file `/etc/systemd/system/openvpn@.service` and run the commands
```
systemctl daemon-reload
systemctl | grep openvpn
```
You will still see the description from `/lib/systemd/system/openvpn@.service` any time.

HenryNe:
Found the problem.
Started bash with debug:
```
export AUTO_INSTALL=y
export IPV6_SUPPORT=y
export ENDPOINT=some.domain.x
export CLIENT=clientname
bash -x ./openvpn-install.sh

... snip ...

Setting up openvpn (2.4.0-6+deb9u3) ...
[ ok ] Restarting virtual private network daemon.:.
Created symlink /etc/systemd/system/multi-user.target.wants/openvpn.service → /lib/systemd/system/openvpn.service.

... snip ...

++ cp /lib/systemd/system/openvpn@.service /etc/systemd/system/openvpn@.service
++ sed -i 's|LimitNPROC|#LimitNPROC|' /etc/systemd/system/openvpn@.service
++ sed -i 's|/etc/openvpn/server|/etc/openvpn|' /etc/systemd/system/openvpn@.service
++ systemctl daemon-reload
++ systemctl restart openvpn@server
Job for openvpn@server.service failed because the control process exited with error code.
See "systemctl status openvpn@server.service" and "journalctl -xe" for details.
++ systemctl enable openvpn@server
Created symlink /etc/systemd/system/multi-user.target.wants/openvpn@server.service → /etc/systemd/system/openvpn@.service.
```
You can see, the `systemctl restart` was before the symlink was set to new file in /etc with the `systemctl enable`.

The order should be

1. systemctl daemon-reload
2. systemctl enable openvpn@server
[...]

Issue #378, opened 2019-01-01
Please Re-Open "Problem running on OpenVZ #129"


StuckUpCreations:
https://github.com/Angristan/OpenVPN-install/issues/129

im still having this problen on debian 9 minimal install and this solution propposed here solves my problem...after commenting out what is suggested here my service started and everything works as expected..

Debian 9 minimal install on OpenVZ

angristan:
Hello, what line are you talking about? It's commented here: https://github.com/angristan/openvpn-install/blob/6e402289bd780efc535beba8cb9d85f766b291bf/openvpn-install.sh#L842

angristan:
Let me know if you find the issue.

parasquid:
Heya, sorry I just got here while debugging my own installation.

```
tristan@quadhost-nc:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.6 LTS
Release:        16.04
Codename:       xenial
```
It would seem like the OS detection is detecting that I have Ubuntu (which I do) but I am running on an OpenVZ host.

Copying and pasting the lines after the `else` in that if branch you linked got everything working.

HenryNe:
The issure again.
```
daemon() failed or unsupported: Resource temporarily unavailable (errno=11)
```
The modifired file `/etc/systemd/system/openvpn@.service` seams not used by Debian 9.
I need to fix the file `/lib/systemd/system/openvpn@.service`, and comment out the **LimitNPROC**, to get it working.

```
# lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 9.9 (stretch)
Release:        9.9
Codename:       stretch

```

Issue #129, opened 2017-11-28
Problem running on OpenVZ


karl-j:
**Error:** 
```Job for openvpn@server.service failed because the control process exited with error code. See "systemctl status openvpn@server.service" and "journalctl -xe" for details.```

Output of `systemctl status openvpn@server.service`:
```
Nov 28 11:58:51 Basic systemd[1]: Starting OpenVPN connection to server...
Nov 28 11:58:51 Basic ovpn-server[30414]: OpenVPN 2.3.10 x86_64-pc-linux-gnu [SSL (OpenSSL)] [LZO] [EPOLL] [PKCS11] [MH] [IPv6] built on 
Nov 28 11:58:51 Basic ovpn-server[30414]: library versions: OpenSSL 1.0.2g  1 Mar 2016, LZO 2.08
Nov 28 11:58:51 Basic ovpn-server[30414]: daemon() failed or unsupported: Resource temporarily unavailable (errno=11)
Nov 28 11:58:51 Basic ovpn-server[30414]: Exiting due to fatal error
Nov 28 11:58:51 Basic systemd[1]: openvpn@server.service: Control process exited, code=exited status=1
Nov 28 11:58:51 Basic systemd[1]: Failed to start OpenVPN connection to server.
Nov 28 11:58:51 Basic systemd[1]: openvpn@server.service: Unit entered failed state.
Nov 28 11:58:51 Basic systemd[1]: openvpn@server.service: Failed with result 'exit-code'.
```

**OS:** Ubuntu 16.04 x64 running on OpenVZ

**Workaround:**
1. comment out the LimitNPROC line in /lib/systemd/system/openvpn@.service
2. systemctl daemon-reload  
    
  
**Credits:**
* https://github.com/lxc/lxd/issues/3336
* https://askubuntu.com/questions/747023/systemd-fails-to-start-openvpn-in-lxd-managed-16-04-container/747146#747146
* https://askubuntu.com/questions/798221/openvpn-in-lxc-failed-to-set-devices-allow

angristan:
Thanks for the issue. There was the same issue with Debian 9.

Could you try this branch? https://github.com/Angristan/OpenVPN-install/blob/systemd-openvz-fix/openvpn-install.sh

michaelfresco:
@Angristan 
Off topic, but the the link to the branch you posted, SOLVED ALL MY PROBLEMS! - I spent 6 hours on getting OpenVPN to work. You are my hero! :)

[...]

PR #584, opened 2020-03-27
Fix Problem running on OpenVZ #583

HenryNe:
On Debian 9 the copy of unit file `/etc/systemd/system/openvpn@.service` has no effect, see #583 .
Same problem as #129 and #378, it can not start on OpenVZ. 

This PR will use a copy of unit file `openvpn-server@.service`  and modyfy it for OpenVZ.

Remember, this PR will change the service name for Debian from `openvpn@server` into `openvpn-server@server`. The same, what is used on fedora|arch|centos.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -894,6 +894,19 @@ verb 3" >> /etc/openvpn/server.conf
 		# This package uses a sysvinit service
 		systemctl enable openvpn
 		systemctl start openvpn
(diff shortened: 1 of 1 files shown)


HenryNe:
Please wait with thist PR.
Afer description edit in `/etc/systemd/system/openvpn@.service` and `systemctl daemon-reload ; systemctl restart` the unit `openvpn@server.service` it is working now.
I will check more.

PR #585, opened 2020-03-27
Fix systemd unit issue on Debian 9

HenryNe:
On Debian 9 the copy of unit file `/etc/systemd/system/openvpn@.service` has no effect, see #583.
Same problem as #129 and #378, unit can not start on OpenVZ. 

`systemctl enable` must run before `systemctl restart`.
So the new link to `/etc/systemd/system/openvpn@.service` was created before `systemctl restart` use it.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -887,8 +887,8 @@ verb 3" >> /etc/openvpn/server.conf
 		fi
 
 		systemctl daemon-reload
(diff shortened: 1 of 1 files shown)


randshell:
If your PR is a work in progress you can create a draft PR next time #584 :)
https://github.blog/2019-02-14-introducing-draft-pull-requests/

PR #591, opened 2020-03-29
Change = conditional to ==

randshell:


angristan:
Please provide a reason for the change...

randshell:
`=` is for variable assignment and `==` for conditions

HenryNe (approved):

angristan:
That's simply no true, otherwise the script would be completely broken :)

https://stackoverflow.com/a/20449556/6945353

> == is a bash-ism, by the way. It's better to use the POSIX =. In bash the two are equivalent, and in plain sh = is the only one guaranteed to work.

randshell:
Well this bash-ism is default on a lot of languages and this script is `#!/bin/bash`. But it's good to know it works because it's a POSIX standard.

HenryNe:
> 
> == is a bash-ism, by the way. It's better to use the POSIX =. In bash the two are equivalent, and in plain sh = is the only one guaranteed to work.

This script only works in bash, because all the `[[ ... ]]` would not work in `sh`.
So, If you use `[[ ... ]]`, then should also use `==`.  _My small comment._

PR #592, opened 2020-03-29
Idempotent in headless mode

### LINKED ISSUE #545: [Patch attached] Make the script idempotent for automated provisioning
Hi! This script has been super-useful to me with a bit of patching to make it more idempotent in headless mode. Maybe others will find it useful as well?

This allows it to plug straight into state-enforcers like Terraform/Ansible/Salt/Puppet/Chef, because it simply works to enforce the desired state of OpenVPN being up and running and a client configuration file being generated.

I didn't change a whole lot actually! Just added a couple checks to make sure random things got saved and OpenVPN and Easy-RSA only got pull down from upstream on the first run.

Gist: [idempotent.patch](https://gist.github.com/jmeubank/9f6a0ffa5d559c5c45b8e00c3d705ed0)
`curl https://gist.githubusercontent.com/jmeubank/9f6a0ffa5d559c5c45b8e00c3d705ed0/raw/8221d8fced9a8e4d0727edebfdef4e1a7e994987/idempotent.patch | git am`
(Review)
`git push origin/master`

jmeubank:
This changeset adjusts the script so that you can run it multiple times with the same input and not have any unexpected changes. This makes it appropriate for "enforcing state", as required by automated provisioners like Puppet, Salt, Chef, or Ansible.

 - Unbound, OpenVPN, easy-rsa, and other dependencies are only installed from upstream if they are not already present. This prevents multiple runs of the script from causing unexpected version upgrades.
 - The easy-rsa CA is only initialized once
 - SERVER_CN and SERVER_NAME are randomly generated once and saved for future reference
 - File append ('>>') is only done strictly after a file is created with '>' (e.g. /etc/sysctl.d/20-openvpn.conf)
 - Clients are only added to easy-rsa once
 - If AUTO_INSTALL == y, then the script operates in install mode and doesn't enter manageMenu

Also, IPv4-only mode has been fixed with a call to curl -4 ifconfig.co.

Fixes: #545

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -662,6 +662,9 @@ function installOpenVPN () {
 			# Install required dependencies and upgrade the system
 			pacman --needed --noconfirm -Syu openvpn iptables openssl wget ca-certificates curl
 		fi
+		if [[ -d /etc/openvpn/easy-rsa/ ]]; then
+			rm -rf /etc/openvpn/easy-rsa/
+		fi
 	fi
 
 	# Find out if the machine uses nogroup or nobody for the permissionless group
@@ -678,7 +681,7 @@ function installOpenVPN () {
 
 	# Install the latest version of easy-rsa from source, if not already
 	# installed.
-	if [[ ! -d /etc/openvpn/easy-rsa-auto/ ]]; then
+	if [[ ! -d /etc/openvpn/easy-rsa/ ]]; then
 		local version="3.0.6"
 		wget -O ~/EasyRSA-unix-v${version}.tgz https://github.com/OpenVPN/easy-rsa/releases/download/v${version}/EasyRSA-unix-v${version}.tgz
 		tar xzf ~/EasyRSA-unix-v${version}.tgz -C ~/
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -662,6 +662,7 @@ function installOpenVPN () {
 			# Install required dependencies and upgrade the system
 			pacman --needed --noconfirm -Syu openvpn iptables openssl wget ca-certificates curl
 		fi
+		# An old version of easy-rsa was available by default in some openvpn packages
 		if [[ -d /etc/openvpn/easy-rsa/ ]]; then
 			rm -rf /etc/openvpn/easy-rsa/
 		fi
@@ -672,11 +673,6 @@ function installOpenVPN () {
 		NOGROUP=nogroup
 	else
 		NOGROUP=nobody
-	fi
-
-	# An old version of easy-rsa was available by default in some openvpn packages
-	if [[ -d /etc/openvpn/easy-rsa/ ]]; then
-		rm -rf /etc/openvpn/easy-rsa/
 	fi
 
 	# Install the latest version of easy-rsa from source, if not already
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -98,6 +98,7 @@ function initialCheck () {
 }
 
 function installUnbound () {
+	# If Unbound isn't installed, install it
 	if [[ ! -e /etc/unbound/unbound.conf ]]; then
 
 		if [[ "$OS" =~ (debian|ubuntu) ]]; then
@@ -137,7 +138,9 @@ prefetch: yes' >> /etc/unbound/unbound.conf
 			# Get root servers list
 			curl -o /etc/unbound/root.hints https://www.internic.net/domain/named.cache
 
-			mv /etc/unbound/unbound.conf /etc/unbound/unbound.conf.old
+			if [[ ! -f /etc/unbound/unbound.conf.old ]]; then
+				mv /etc/unbound/unbound.conf /etc/unbound/unbound.conf.old
+			fi
 
 			echo 'server:
 	use-syslog: yes
@@ -596,8 +599,12 @@ function installOpenVPN () {
 		PASS=${PASS:-1}
 		CONTINUE=${CONTINUE:-y}
 
-		# Behind NAT, we'll default to the publicly reachable IPv4.
-		PUBLIC_IPV4=$(curl ifconfig.co)
+		# Behind NAT, we'll default to the publicly reachable IPv4/IPv6.
+		if [[ $IPV6_SUPPORT == "y" ]]; then
+			PUBLIC_IPV4=$(curl ifconfig.co)
+		else
+			PUBLIC_IPV4=$(curl -4 ifconfig.co)
+		fi
 		ENDPOINT=${ENDPOINT:-$PUBLIC_IPV4}
 	fi
 
@@ -623,33 +630,38 @@ function installOpenVPN () {
                 fi
         fi
 
-	if [[ "$OS" =~ (debian|ubuntu) ]]; then
-		apt-get update
-		apt-get -y install ca-certificates gnupg
-		# We add the OpenVPN repo to get the latest version.
-		if [[ "$VERSION_ID" == "8" ]]; then
-			echo "deb http://build.openvpn.net/debian/openvpn/stable jessie main" > /etc/apt/sources.list.d/openvpn.list
-			wget -O - https://swupdate.openvpn.net/repos/repo-public.gpg | apt-key add -
+	# If OpenVPN isn't installed yet, install it. This script is more-or-less
+	# idempotent on multiple runs, but will only install OpenVPN from upstream
+	# the first time.
+	if [[ ! -e /etc/openvpn/server.conf ]]; then
+		if [[ "$OS" =~ (debian|ubuntu) ]]; then
 			apt-get update
+			apt-get -y install ca-certificates gnupg
+			# We add the OpenVPN repo to get the latest version.
+			if [[ "$VERSION_ID" = "8" ]]; then
+				echo "deb http://build.openvpn.net/debian/openvpn/stable jessie main" > /etc/apt/sources.list.d/openvpn.list
+				wget -O - https://swupdate.openvpn.net/repos/repo-public.gpg | apt-key add -
+				apt-get update
+			fi
+			if [[ "$VERSION_ID" = "16.04" ]]; then
+				echo "deb http://build.openvpn.net/debian/openvpn/stable xenial main" > /etc/apt/sources.list.d/openvpn.list
+				wget -O - https://swupdate.openvpn.net/repos/repo-public.gpg | apt-key add -
+				apt-get update
+			fi
+			# Ubuntu > 16.04 and Debian > 8 have OpenVPN >= 2.4 without the need of a third party repository.
+			apt-get install -y openvpn iptables openssl wget ca-certificates curl
+		elif [[ "$OS" = 'centos' ]]; then
+			yum install -y epel-release
+			yum install -y openvpn iptables openssl wget ca-certificates curl tar 'policycoreutils-python*'
+		elif [[ "$OS" = 'amzn' ]]; then
+			amazon-linux-extras install -y epel
+			yum install -y openvpn iptables openssl wget ca-certificates curl
+		elif [[ "$OS" = 'fedora' ]]; then
+			dnf install -y openvpn iptables openssl wget ca-certificates curl
+		elif [[ "$OS" = 'arch' ]]; then
+			# Install required dependencies and upgrade the system
+			pacman --needed --noconfirm -Syu openvpn iptables openssl wget ca-certificates curl
 		fi
-		if [[ "$VERSION_ID" == "16.04" ]]; then
-			echo "deb http://build.openvpn.net/debian/openvpn/stable xenial main" > /etc/apt/sources.list.d/openvpn.list
-			wget -O - https://swupdate.openvpn.net/repos/repo-public.gpg | apt-key add -
-			apt-get update
-		fi
-		# Ubuntu > 16.04 and Debian > 8 have OpenVPN >= 2.4 without the need of a third party repository.
-		apt-get install -y openvpn iptables openssl wget ca-certificates curl
-	elif [[ "$OS" == 'centos' ]]; then
-		yum install -y epel-release
-		yum install -y openvpn iptables openssl wget ca-certificates curl tar 'policycoreutils-python*'
-	elif [[ "$OS" == 'amzn' ]]; then
-		amazon-linux-extras install -y epel
-		yum install -y openvpn iptables openssl wget ca-certificates curl
-	elif [[ "$OS" == 'fedora' ]]; then
-		dnf install -y openvpn iptables openssl wget ca-certificates curl
-	elif [[ "$OS" == 'arch' ]]; then
-		# Install required dependencies and upgrade the system
-		pacman --needed --noconfirm -Syu openvpn iptables openssl wget ca-certificates curl
 	fi
 
 	# Find out if the machine uses nogroup or nobody for the permissionless group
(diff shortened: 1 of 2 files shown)


openvpn-install.sh
```diff
@@ -596,8 +599,12 @@ function installOpenVPN () {
...
 
-		# Behind NAT, we'll default to the publicly reachable IPv4.
-		PUBLIC_IPV4=$(curl ifconfig.co)
+		# Behind NAT, we'll default to the publicly reachable IPv4/IPv6.
+		if [[ $IPV6_SUPPORT == "y" ]]; then
+			PUBLIC_IPV4=$(curl ifconfig.co)
+		else
+			PUBLIC_IPV4=$(curl -4 ifconfig.co)
+		fi
 		ENDPOINT=${ENDPOINT:-$PUBLIC_IPV4}
```
randshell: @angristan are we sure we can change this behavior to default to an IPv6 endpoint?

randshell:
I only have a question. I don't understand the change to `easy-rsa-auto`, we didn't have conflicts for now.

In the list of provisioning you can add Ansible too.

Thanks for contributing!

jmeubank:
> I only have a question. I don't understand the change to `easy-rsa-auto`, we didn't have conflicts for now.

[This line](https://github.com/angristan/openvpn-install/blob/44105eb06068ef95ff3217924ee615e8757f7516/openvpn-install.sh#L662) mentions deleting a pre-existing 'easy-rsa' folder that came with older OpenVPN installations. On systems where this folder may pre-exist, it's safer to use a different folder name for the new incoming easy-rsa. Otherwise, it'd be hard to tell on repeated runs of openvpn-install.sh whether the contents of the folder are what we want them to be.

> In the list of provisioning you can add Ansible too.

Done! [e350220](https://github.com/angristan/openvpn-install/pull/592/commits/e35022065b42e46baac7a88de5c0c0345d4a1ec3)

randshell:
> On systems where this folder may pre-exist, it's safer to use a different folder name for the new incoming easy-rsa

But you didn't delete the line `rm -rf /etc/openvpn/easy-rsa/` so `easy-rsa` will be clean after installation anyway.

jmeubank:
> > On systems where this folder may pre-exist, it's safer to use a different folder name for the new incoming easy-rsa
> 
> But you didn't delete the line `rm -rf /etc/openvpn/easy-rsa/` so `easy-rsa` will be clean after installation anyway.

But on the next run of the script, easy-rsa-auto will be present. If I kept the original naming convention, easy-rsa would be present and would be deleted and recreated each time.

The goal is for the desired version of easy-rsa (currently 3.0.6) to be installed once by this script. If an older version that was packaged with OpenVPN in the distro's repos is present (in easy-rsa), it should be deleted. Then, subsequent runs of this script shouldn't re-download and recreate it.

openvpn-install.sh
```diff
@@ -623,33 +630,38 @@ function installOpenVPN () {
...
-		yum install -y openvpn iptables openssl wget ca-certificates curl tar 'policycoreutils-python*'
-	elif [[ "$OS" = 'amzn' ]]; then
-		amazon-linux-extras install -y epel
-		yum install -y openvpn iptables openssl wget ca-certificates curl
-	elif [[ "$OS" = 'fedora' ]]; then
-		dnf install -y openvpn iptables openssl wget ca-certificates curl
-	elif [[ "$OS" = 'arch' ]]; then
-		# Install required dependencies and upgrade the system
-		pacman --needed --noconfirm -Syu openvpn iptables openssl wget ca-certificates curl
 	fi
```
randshell, line 669:
```suggestion
		if [[ -d /etc/openvpn/easy-rsa/ ]]; then
			rm -rf /etc/openvpn/easy-rsa/
		fi
	fi
```

openvpn-install.sh
```diff
@@ -664,60 +676,72 @@ function installOpenVPN () {
...
-	# Install the latest version of easy-rsa from source
-	local version="3.0.6"
-	wget -O ~/EasyRSA-unix-v${version}.tgz https://github.com/OpenVPN/easy-rsa/releases/download/v${version}/EasyRSA-unix-v${version}.tgz
-	tar xzf ~/EasyRSA-unix-v${version}.tgz -C ~/
-	mv ~/EasyRSA-v${version} /etc/openvpn/easy-rsa
-	chown -R root:root /etc/openvpn/easy-rsa/
-	rm -f ~/EasyRSA-unix-v${version}.tgz
+	# Install the latest version of easy-rsa from source, if not already
+	# installed.
+	if [[ ! -d /etc/openvpn/easy-rsa-auto/ ]]; then
```
```suggestion
	if [[ ! -d /etc/openvpn/easy-rsa/ ]]; then
```

openvpn-install.sh
```diff
@@ -664,60 +676,72 @@ function installOpenVPN () {
...
+		local version="3.0.6"
+		wget -O ~/EasyRSA-unix-v${version}.tgz https://github.com/OpenVPN/easy-rsa/releases/download/v${version}/EasyRSA-unix-v${version}.tgz
+		tar xzf ~/EasyRSA-unix-v${version}.tgz -C ~/
+		mkdir -p /etc/openvpn/easy-rsa-auto
+		mv ~/EasyRSA-v${version}/* /etc/openvpn/easy-rsa-auto/
+		chown -R root:root /etc/openvpn/easy-rsa-auto/
```
```suggestion
		mkdir -p /etc/openvpn/easy-rsa
		mv ~/EasyRSA-v${version}/* /etc/openvpn/easy-rsa/
		chown -R root:root /etc/openvpn/easy-rsa/
```

openvpn-install.sh
```diff
@@ -664,60 +676,72 @@ function installOpenVPN () {
...
+		rm -f ~/EasyRSA-unix-v${version}.tgz
+
+		cd /etc/openvpn/easy-rsa-auto/ || return
```
```suggestion
		cd /etc/openvpn/easy-rsa/ || return
```

openvpn-install.sh
```diff
@@ -664,60 +676,72 @@ function installOpenVPN () {
...
+			;;
+			2)
+				# Generate tls-auth key
+				openvpn --genkey --secret /etc/openvpn/tls-auth.key
+			;;
+		esac
+	else
+		# If easy-rsa is already installed, grab the generated SERVER_NAME
+		# for client configs
+		cd /etc/openvpn/easy-rsa-auto/ || return
```
```suggestion
		cd /etc/openvpn/easy-rsa/ || return
```

openvpn-install.sh
```diff
@@ -664,60 +676,72 @@ function installOpenVPN () {
...
+		SERVER_NAME=$(cat SERVER_NAME_GENERATED)
+	fi
 
 	# Move all the generated files
-	cp pki/ca.crt pki/private/ca.key "pki/issued/$SERVER_NAME.crt" "pki/private/$SERVER_NAME.key" /etc/openvpn/easy-rsa/pki/crl.pem /etc/openvpn
+	cp pki/ca.crt pki/private/ca.key "pki/issued/$SERVER_NAME.crt" "pki/private/$SERVER_NAME.key" /etc/openvpn/easy-rsa-auto/pki/crl.pem /etc/openvpn
```
```suggestion
	cp pki/ca.crt pki/private/ca.key "pki/issued/$SERVER_NAME.crt" "pki/private/$SERVER_NAME.key" /etc/openvpn/easy-rsa/pki/crl.pem /etc/openvpn
```

openvpn-install.sh
```diff
@@ -1026,16 +1050,23 @@ function newClient () {
...
-	case $PASS in
-		1)
-			./easyrsa build-client-full "$CLIENT" nopass
-		;;
-		2)
-		echo "⚠️ You will be asked for the client password below ⚠️"
-			./easyrsa build-client-full "$CLIENT"
-		;;
-	esac
+	CLIENTEXISTS=$(tail -n +2 /etc/openvpn/easy-rsa-auto/pki/index.txt | grep -c -E "/CN=$CLIENT\$")
```
```suggestion
	CLIENTEXISTS=$(tail -n +2 /etc/openvpn/easy-rsa/pki/index.txt | grep -c -E "/CN=$CLIENT\$")
```

openvpn-install.sh
```diff
@@ -1026,16 +1050,23 @@ function newClient () {
...
+	if [[ "$CLIENTEXISTS" = '1' ]]; then
+		echo ""
+		echo "The specified client CN was found in easy-rsa."
+	else
+		cd /etc/openvpn/easy-rsa-auto/ || return
```
```suggestion
		cd /etc/openvpn/easy-rsa/ || return
```

openvpn-install.sh
```diff
@@ -1057,15 +1088,15 @@ function newClient () {
 	cp /etc/openvpn/client-template.txt "$homeDir/$CLIENT.ovpn"
 	{
 		echo "<ca>"
-		cat "/etc/openvpn/easy-rsa/pki/ca.crt"
+		cat "/etc/openvpn/easy-rsa-auto/pki/ca.crt"
```
```suggestion
		cat "/etc/openvpn/easy-rsa/pki/ca.crt"
```

openvpn-install.sh
```diff
@@ -1057,15 +1088,15 @@ function newClient () {
...
 		echo "</ca>"
 
 		echo "<cert>"
-		awk '/BEGIN/,/END/' "/etc/openvpn/easy-rsa/pki/issued/$CLIENT.crt"
+		awk '/BEGIN/,/END/' "/etc/openvpn/easy-rsa-auto/pki/issued/$CLIENT.crt"
```
```suggestion
		awk '/BEGIN/,/END/' "/etc/openvpn/easy-rsa/pki/issued/$CLIENT.crt"
```

openvpn-install.sh
```diff
@@ -1057,15 +1088,15 @@ function newClient () {
...
 		echo "</cert>"
 
 		echo "<key>"
-		cat "/etc/openvpn/easy-rsa/pki/private/$CLIENT.key"
+		cat "/etc/openvpn/easy-rsa-auto/pki/private/$CLIENT.key"
```
```suggestion
		cat "/etc/openvpn/easy-rsa/pki/private/$CLIENT.key"
```

openvpn-install.sh
```diff
@@ -1084,14 +1115,14 @@ function newClient () {
...
-	echo "Client $CLIENT added, the configuration file is available at $homeDir/$CLIENT.ovpn."
+	echo "The configuration file has been written to $homeDir/$CLIENT.ovpn."
 	echo "Download the .ovpn file and import it in your OpenVPN client."
 
 	exit 0
 }
 
 function revokeClient () {
-	NUMBEROFCLIENTS=$(tail -n +2 /etc/openvpn/easy-rsa/pki/index.txt | grep -c "^V")
+	NUMBEROFCLIENTS=$(tail -n +2 /etc/openvpn/easy-rsa-auto/pki/index.txt | grep -c "^V")
```
```suggestion
	NUMBEROFCLIENTS=$(tail -n +2 /etc/openvpn/easy-rsa/pki/index.txt | grep -c "^V")
```

openvpn-install.sh
```diff
@@ -1100,23 +1131,23 @@ function revokeClient () {
 
 	echo ""
 	echo "Select the existing client certificate you want to revoke"
-	tail -n +2 /etc/openvpn/easy-rsa/pki/index.txt | grep "^V" | cut -d '=' -f 2 | nl -s ') '
+	tail -n +2 /etc/openvpn/easy-rsa-auto/pki/index.txt | grep "^V" | cut -d '=' -f 2 | nl -s ') '
```
```suggestion
	tail -n +2 /etc/openvpn/easy-rsa/pki/index.txt | grep "^V" | cut -d '=' -f 2 | nl -s ') '
```

HenryNe:
> 
> 
> > > On systems where this folder may pre-exist, it's safer to use a different folder name for the new incoming easy-rsa
> > 
> > 
> > But you didn't delete the line `rm -rf /etc/openvpn/easy-rsa/` so `easy-rsa` will be clean after installation anyway.
> 
> But on the next run of the script, easy-rsa-auto will be present. If I kept the original naming convention, easy-rsa would be present and would be deleted and recreated each time.
> 
> The goal is for the desired version of easy-rsa (currently 3.0.6) to be installed once by this script. If an older version that was packaged with OpenVPN in the distro's repos is present (in easy-rsa), it should be deleted. Then, subsequent runs of this script shouldn't re-download and recreate it.

An other question is, what does an update of the distro, for example if  the distro will be update easy-rsa?

openvpn-install.sh
```diff
@@ -1100,23 +1131,23 @@ function revokeClient () {
...
 	if [[ "$NUMBEROFCLIENTS" = '1' ]]; then
 		read -rp "Select one client [1]: " CLIENTNUMBER
 	else
 		read -rp "Select one client [1-$NUMBEROFCLIENTS]: " CLIENTNUMBER
 	fi
 
-	CLIENT=$(tail -n +2 /etc/openvpn/easy-rsa/pki/index.txt | grep "^V" | cut -d '=' -f 2 | sed -n "$CLIENTNUMBER"p)
-	cd /etc/openvpn/easy-rsa/ || return
+	CLIENT=$(tail -n +2 /etc/openvpn/easy-rsa-auto/pki/index.txt | grep "^V" | cut -d '=' -f 2 | sed -n "$CLIENTNUMBER"p)
+	cd /etc/openvpn/easy-rsa-auto/ || return
```
```suggestion
	CLIENT=$(tail -n +2 /etc/openvpn/easy-rsa/pki/index.txt | grep "^V" | cut -d '=' -f 2 | sed -n "$CLIENTNUMBER"p)
	cd /etc/openvpn/easy-rsa/ || return
```

openvpn-install.sh
```diff
@@ -1100,23 +1131,23 @@ function revokeClient () {
...
 	./easyrsa --batch revoke "$CLIENT"
 	EASYRSA_CRL_DAYS=3650 ./easyrsa gen-crl
 	# Cleanup
 	rm -f "pki/reqs/$CLIENT.req"
 	rm -f "pki/private/$CLIENT.key"
 	rm -f "pki/issued/$CLIENT.crt"
 	rm -f /etc/openvpn/crl.pem
-	cp /etc/openvpn/easy-rsa/pki/crl.pem /etc/openvpn/crl.pem
+	cp /etc/openvpn/easy-rsa-auto/pki/crl.pem /etc/openvpn/crl.pem
```
```suggestion
	cp /etc/openvpn/easy-rsa/pki/crl.pem /etc/openvpn/crl.pem
```

randshell (changes requested):
> subsequent runs of this script shouldn't re-download and recreate it.

Now I get it. It would happen when you try to run the command to install it again but it's already installed.

I think we can avoid changing the name of the `easy-rsa` directory to `easy-rsa-auto` if we move `rm -rf /etc/openvpn/easy-rsa/` to be inside `if [[ ! -e /etc/openvpn/server.conf ]]; then`. Can you let me know if you agree please?

jmeubank:
> An other question is, what does an update of the distro, for example if the distro will be update easy-rsa?

If the distro updates easy-rsa, it'll go in a separate directory and won't bother us.

If the distro overwrites the OpenVPN config referencing easy-rsa, we have a bigger problem, but re-running openvpn-install.sh will fix it. :)

jmeubank:
> I think we can avoid changing the name of the `easy-rsa` directory to `easy-rsa-auto` if we move `rm -rf /etc/openvpn/easy-rsa/` to be inside `if [[ ! -e /etc/openvpn/server.conf ]]; then`. Can you let me know if you agree please?

Agree, that's much cleaner!

randshell (approved):
Thanks!

openvpn-install.sh
angristan: I think most people would still prefer to connect to an IPv4 endpoint...

angristan:
Seems great guys, thanks 👍

openvpn-install.sh
HenryNe: I prefer also an IPv4 endpoint.
An IPv6 endpoint should only a fallback, if IPv4 is not usable.

For my understand the IPV6_SUPPORT is a define to support IPv6 and NAT inside the VPN.
For a dual stack endpoint I enable IPV6_SUPPORT and put a FDNS name as endpoint.

jmeubank: curl ifconfig.co in the current version returns an IPv6 address on my servers, which is why I made this change - although my servers are IPv6 enabled, I'm disabling IPv6 in OpenVPN (and forcing an IPv6 blackhole with route-ipv6 2000::/3 dead::beef). I could set ENDPOINT separately from the script, so this isn't a make-or-break change for me, but it was an easy fix. And for dynamically assigned IPs, it's nice to have.

randshell: @jmeubank Can you please change it to `https://ifconfig.co` to use the https version? You are already working on this line so I'm not going to open a new PR.

jmeubank: No prob - done.

openvpn-install.sh
```diff
@@ -596,8 +599,12 @@ function installOpenVPN () {
...
 
-		# Behind NAT, we'll default to the publicly reachable IPv4.
-		PUBLIC_IPV4=$(curl ifconfig.co)
+		# Behind NAT, we'll default to the publicly reachable IPv4/IPv6.
+		if [[ $IPV6_SUPPORT == "y" ]]; then
+			PUBLIC_IPV4=$(curl https://ifconfig.co)
+		else
+			PUBLIC_IPV4=$(curl -4 https://ifconfig.co)
+		fi
 		ENDPOINT=${ENDPOINT:-$PUBLIC_IPV4}
```
angristan: I guess we can rename `PUBLIC_IPV4` to `PUBLIC_IP` now 🙂

angristan (changes requested):

openvpn-install.sh
jmeubank: Easy, done. :)

jmeubank:
I did my habitual pull rebase and dirtied up this PR, sorry. Will reset and clean it up.

jmeubank:
Okay, it's clean again, let me know if you need me to squash any of the commits. :)

angristan:
Don't worry I'll squash them. Thanks again, I just have to find some time to test it and merge it!

openvpn-install.sh
```diff
@@ -664,57 +679,69 @@ function installOpenVPN () {
...
-	# Install the latest version of easy-rsa from source
-	local version="3.0.6"
-	wget -O ~/EasyRSA-unix-v${version}.tgz https://github.com/OpenVPN/easy-rsa/releases/download/v${version}/EasyRSA-unix-v${version}.tgz
-	tar xzf ~/EasyRSA-unix-v${version}.tgz -C ~/
-	mv ~/EasyRSA-v${version} /etc/openvpn/easy-rsa
-	chown -R root:root /etc/openvpn/easy-rsa/
-	rm -f ~/EasyRSA-unix-v${version}.tgz
+	# Install the latest version of easy-rsa from source, if not already
+	# installed.
+	if [[ ! -d /etc/openvpn/easy-rsa/ ]]; then
```
angristan, line 680: This is ineffective because we delete the folder right before :)

angristan (changes requested):

openvpn-install.sh
randshell: I suggested to @jmeubank to move the deletion code of easy-rsa that comes with openvpn to be after the openvpn installation to avoid renaming to easy-rsa-auto https://github.com/angristan/openvpn-install/pull/592#pullrequestreview-383455081

@angristan You are right, I missed the fact that we didn't delete the old remove code so now easy-rsa gets removed every time anyway.
https://github.com/angristan/openvpn-install/blob/5805cc03b2a45fb9dd859b60a5e6fe30d2cf28a2/openvpn-install.sh#L677-L680
This should be removed.

randshell (changes requested):
Thanks @angristan

openvpn-install.sh
jmeubank: Yeah. I think I screwed that up with my rebase. Fixing!

angristan (approved):

angristan:
Thanks a lot everyone!

PR #594, opened 2020-03-30
Add `/usr/sbin` to Debian `$PATH`

### LINKED ISSUE #523: Issue with Debian 10 <tls-crypt> empty
Hi there,

first, thanks alot for your script, been using it several years now.

Now ive got issues using your script with debian 10.

When creating a new user the <tls-crypt> part is empty, so its not possible to import it in any openvpn client.

I think it has something todo with the line:
openvpn --genkey --secret /etc/openvpn/tls-crypt.key
in your script.

any workarounds here?

### LINKED ISSUE #593: Complains if /usr/sbin isn't in root's PATH
Hi! Nice script! :D

Just one minor problem: if /usr/sbin isn't in root's PATH, launching "openvpn" and "systemctl" fail ("Command not found"), but it seems to work the same.

Thanks!

    M

randshell:
Fix #523, fix #593

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -35,6 +35,8 @@ function checkOS () {
 					exit 1
 				fi
 			fi
+			# Fix $PATH on Debian
+			PATH=${PATH}:/usr/sbin
 		elif [[ "$ID" == "ubuntu" ]];then
 			OS="ubuntu"
 			if [[ ! $VERSION_ID =~ (16.04|18.04|19.04) ]]; then


HenryNe:
This would not enouth. There are some more paths missing for a user, for example `/sbin` for `/sbin/sysctl`, used in openvpn-install.sh#L864

```
# which sysctl
/sbin/sysctl
```

randshell:
I tested on Debian 10 where the issue was reported and `sysctl` is at `/usr/sbin/sysctl`.

Where do you get that `sysctl` is at `/sbin/sysctl` and not already in the `$PATH`?

HenryNe:
A default Debian 9 installation on vServer.
```
user@server:~$ sysctl
-bash: sysctl: command not found

user@server:~$ id
uid=1000(user) gid=1000(user) groups=1000(user)

user@server:~$ echo $PATH
/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games

user@server:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 9.0 (stretch)
Release:        9.0
Codename:       stretch
```

HenryNe:
On Debian 9 the current script works also with su. The PATH for root is complete after su:
```
user@server:~$ id
uid=1000(user) gid=1000(user) groups=1000(user)
user@server:~$ echo $PATH
/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
user@server:~$ su
Password:
root@server:/home/user# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

HenryNe:
In generally, I approve your PR. It makes it more resistant.

randshell:
For the currently supported distros it shouldn't be necessary to add `/sbin`. There is https://github.com/pivpn/pivpn/commit/2eb11cad0ced98482eaeccc042aa8d5f291ffb57 but it means supporting other distros and no one opened an issue for it.

@HenryNe thanks for investigating this.

angristan:
https://github.com/angristan/openvpn-install/issues/593#issuecomment-607213175

PR #599, opened 2020-03-30
Get DNS-IPv6 from current system resolvers

HenryNe:
Allow IPv4 and IPv6 to get from resolv.conf.
Filter out commets by grepping "nameserver" from beginning of line. No additional comment filter need.
Replace multiple grep with single sed.


angristan:
Shouldn't we get IPv6 resolvers only if IPv6 support is enabled?

HenryNe:
> 
> 
> Shouldn't we get IPv6 resolvers only if IPv6 support is enabled?

Yes, I was also thinging about.
I will change the PR.

openvpn-install.sh
```diff
@@ -754,8 +754,10 @@ ifconfig-pool-persist ipp.txt" >> /etc/openvpn/server.conf
 				RESOLVCONF='/etc/resolv.conf'
 			fi
 			# Obtain the resolvers from resolv.conf and use them for OpenVPN
-			grep -v '#' $RESOLVCONF | grep 'nameserver' | grep -E -o '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}' | while read -r line; do
-				echo "push \"dhcp-option DNS $line\"" >> /etc/openvpn/server.conf
+			sed -ne 's/^nameserver[[:space:]]\+\([^[:space:]]\+\).*$/\1/p' $RESOLVCONF | while read -r line; do
+				if [[ "$line" =~ ^[0-9.]*$ ]] || [[ "$IPV6_SUPPORT" == 'y' ]]; then
```
angristan: So now, IPv4 resolvers are ignored?

randshell: Doesn't it mean if `$line` is IPv4 write it or if it's not IPv4 write it if  `$IPV6_SUPPORT" == 'y'`?
I tested and it works correctly.

angristan: Oops, I misread! That condition seems odd to me though. 

_If there's something behind `nameserver`_, _write it if it's kinda like an IPv4_ (the regex is not strict) _or write if it's anything but IPv6 is enabled_

It will probably not fail because the resolv.conf file is probably valid, but nonetheless...

It should be: `it's a valid IPv4 || (it's a valid ipv6 && ipv6 support is enabled)`

angristan (changes requested):

openvpn-install.sh
randshell: Yes, you're right to check if it's IPv6 to be sure.

HenryNe: Complete IPv6 checks would be very difficould: https://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses

We can add such a simple check. But this make the line difficould to, because the OR and AND.
Hera is one variant, but I'm not shure, it's save in all bash versions?
For this you need to known, `&&` actually has higher precedence than `||` :
```
if [[ "$line" =~ ^[0-9.]*$ || "$line" =~ ^[0-9A-Fa-f:.]*$ && "$IPV6_SUPPORT" == 'y' ]]; then
     echo  ..
fi
```
A liddle saver would be:
```
if [[ "$line" =~ ^[0-9.]*$ ]] || [[ "$line" =~ ^[0-9A-Fa-f:.]*$ && "$IPV6_SUPPORT" == 'y' ]]; then
     echo  ..
fi
```

This would be an other variant, that is saver for all bash versions:
```
if [[ "$line" =~ ^[0-9.]*$ ]]; then
     echo  ..
elif [[ "$line" =~ ^[0-9A-Fa-f:.]*$ ]] && [[ "$IPV6_SUPPORT" == 'y' ]]; then
     echo  ..
fi
```

For me all extra checks are overdriven. recolve.conf should have only valid entries: IPv4 address or IPv6 address, one per line, nothing else.

HenryNe:
By the way: `01002004010` and `134744072` are also valid IPv4 addresses, also without dots. Try it: https://01002004010/ or `ping 134744072` it is google DNS 8.8.8.8

angristan (approved):
Works fine 👍

PR #600, opened 2020-03-30
Add DNS IPv6 for self-hosted DNS Resolver

HenryNe:


randshell (changes requested):
Unbound is configured to listen only on IPv4 
https://github.com/angristan/openvpn-install/blob/44105eb06068ef95ff3217924ee615e8757f7516/openvpn-install.sh#L107
https://github.com/angristan/openvpn-install/blob/44105eb06068ef95ff3217924ee615e8757f7516/openvpn-install.sh#L118

randshell:
You need to append `interface: fd42:42:42:42::1`, `access-control: fd42:42:42:42::/112 allow` and `private-address:  fd42:42:42:42::/112` to unbound config.

HenryNe:
@randomshell thanks for the watch. unbound.conf changed, .
Debian 9 as server tested successfully.

On the client side have tested with an explizit IPv6 command:

```
$ host -v -6 www.google.com
...
;; ANSWER SECTION:
www.google.com.         229     IN      AAAA    2a00:1450:4001:81d::2004
...
Received 82 bytes from fd42:42:42:42::1#53 in 91 ms
```

openvpn-install.sh
```diff
@@ -181,13 +188,18 @@ hide-version: yes
...
 private-address: 172.16.0.0/12
 private-address: 192.168.0.0/16
 private-address: 169.254.0.0/16
 private-address: fd00::/8
 private-address: fe80::/10
 private-address: 127.0.0.0/8
 private-address: ::ffff:0:0/96' > /etc/unbound/openvpn.conf
+		if [[ "$IPV6_SUPPORT" == 'y' ]]; then
+			echo 'interface: fd42:42:42:42::1
+access-control: fd42:42:42:42::/112 allow' >> /etc/unbound/unbound.conf
```
randomshell: Fix unbound configuration when already installed
```suggestion
access-control: fd42:42:42:42::/112 allow' >> /etc/unbound/openvpn.conf
```

randshell: Fix unbound configuration path when already installed
```suggestion
access-control: fd42:42:42:42::/112 allow' >> /etc/unbound/openvpn.conf
```

randshell (changes requested):

randshell (approved):
Thanks!

angristan:
Thanks! While I'm not against this per se, what is the point of a dual-stack **local** resolver?

HenryNe:
> 
> 
> Thanks! While I'm not against this per se, what is the point of a dual-stack **local** resolver?

My server is a IPv6 only system. So, in my case VPN is usable only via IPv6. This server has no IPv4 default route. This server give not a real dual stack VPN. Only the IPv6 part is usable. That's why I fall in some pitfalls, where IPv6 is not full usable. - I would witsh to disable IPv4. But this would be an other feature.

In an other case, Vagrant have no IPv6 support. With VPN connection to the IPv6 server I have full IPv6 connections in Vagrant, and can test IPv6 only systms in Vagrant.

angristan:
OK. So with this PR Unbound will be usable on:

- IPv4-only systems
- Dual-stack
- IPv6-only systems

Correct? If that's the case, then I'm all for it 👍

HenryNe:
Yes, and explicit **yes** for IPv6-only.

angristan (changes requested):
OK with this, just need a rebase against master 👍

angristan (approved):

PR #601, opened 2020-03-30
Add IPv6 INPUT on incomming port

HenryNe:
This adds and remove ip6tables for incomming traffic over IPv6 on the configured port (1194).


randshell (approved):

angristan:
Thanks. When I added IPv6 support I was still considering users would connect to an IPv4 endpoint

HenryNe:
My home provider have only IPv6. All IPv4 connections goes over DS-Lite. That is sometimes not nice. So I'm searching for faster IPv6 connetions.

I have seen, that VPN is listen on IPv6, I used, and found, that iptables was not the same as in IPv4.

PR #604, opened 2020-03-31
one unbound config for all OS, fix #602

### LINKED ISSUE #602: unbound not cleanly removed
After cleaning and re-installing openvn with unbound (DNS=2) unbound configuration hangs in an endless loop without changeing unbound config, or with double config.

1. Start without openvpn installed, and without unbound installed
2. Run install, choose **DNS=2**, the all the rest defaults:
```
APPROVE_INSTALL=y \
APPROVE_IP=y \
IPV6_SUPPORT=n \
PORT_CHOICE=1 \
PROTOCOL_CHOICE=1 \
DNS=2 \
COMPRESSION_ENABLED=n \
CUSTOMIZE_ENC=n \
CLIENT=clientname \
PASS=1 \
./openvpn-install.sh
```
3. Remove openvpn, start `./openvpn-install.sh` again and choose **menu 3**. There is no question about removing unbound.
4. Run `openvpn-install.sh` to install it again, select **DNS=2**, for all others questions Enter default
5. It would detect "Unbound is already installed."
6. If you say "n" for "Apply configuration changes to Unbound?", you came again to question DNS. Select 2 for DNS. This goes endless.
7. Choose "y" for "Apply configuration changes to Unbound?"
8. Enter defaults for the rest

As result the VPN configuration in `/etc/unbound/unbound.conf` exist two times. Ones in `/etc/unbound/unbound.conf` and second via the
```
include: /etc/unbound/openvpn.conf
```
For now unbound not works and failed with
```
unbound[15550:0] error: can't bind socket: Address already in use for 10.8.0.1
unbound[15550:0] fatal error: could not open ports
```
I suggest do not put values into /etc/unbound/unbound.conf, only use one config in all cases, for example `/etc/unbound/openvpn.conf`, or better `/etc/unbound/unbound.conf.d/openvpn.conf` and detect the include, before add the include again.

OS: Debian 9
unbound 1.6.0-3+deb9u2

The same error in #427, also Debian 9.

HenryNe:
Fix for #602
Create config `/etc/unbound/unbound.conf.d/openvpn.conf`,
include this config into `/etc/unbound/unbound.conf`,
do it for all OS.

On uninstall simple remove the include, if if was injected by openvpn-install.

_edit:_
tested on Debian 9 and Arch Linux 1.4


openvpn-install.sh
```diff
@@ -1128,8 +1096,8 @@ function revokeClient () {
 
 function removeUnbound () {
 	# Remove OpenVPN-related config
-	sed -i 's|include: \/etc\/unbound\/openvpn.conf||' /etc/unbound/unbound.conf
-	rm /etc/unbound/openvpn.conf
+	sed -i '/include .*\/openvpn.conf/d' /etc/unbound/unbound.conf
```
randshell: Can we use `/etc/unbound/unbound.conf.d/openvpn.conf` instead of wildcard?

openvpn-install.sh
```diff
@@ -146,48 +124,38 @@ prefetch: yes' >> /etc/unbound/unbound.conf
...
-	access-control: 10.8.0.1/24 allow
 	port: 53
 	num-threads: 2
-	use-caps-for-id: yes
 	harden-glue: yes
-	hide-identity: yes
-	hide-version: yes
-	qname-minimisation: yes
-	prefetch: yes' > /etc/unbound/unbound.conf
+	qname-minimisation: yes' >/etc/unbound/unbound.conf
```
randshell: `harden-glue: yes` and `qname-minimisation: yes` are for all OS not only Arch. On `unbound.conf` manual they default to `yes` anyway.

openvpn-install.sh
```diff
@@ -146,48 +124,38 @@ prefetch: yes' >> /etc/unbound/unbound.conf
...
 access-control: 10.8.0.1/24 allow
 hide-identity: yes
 hide-version: yes
 use-caps-for-id: yes
-prefetch: yes
-private-address: 10.0.0.0/8
+prefetch: yes' >/etc/unbound/unbound.conf.d/openvpn.conf
+
+	if [[ ! "$OS" =~ (fedora|centos|amzn) ]];then
+		# DNS Rebinding fix
```
randshell: DNS rebinding fix is for all OS not only fedora, centos or amzn

randshell (changes requested):

openvpn-install.sh
HenryNe: Yes. It's only longer, would not find the old location `/etc/unbound/openvpn.conf`, and needs lot of backslashes  ;-)
`
sed -i '/include \/etc\/unbound\/unbound.conf.d\/openvpn.conf/d' /etc/unbound/unbound.conf
`

openvpn-install.sh
HenryNe: OK. Have copied from old, and I would not change function here.

openvpn-install.sh
randshell: I think it's clearer to specify the full path but I'm fine anyway.

randshell:
@angristan I think in Arch we can remove the following code because they are defaults.
https://github.com/angristan/openvpn-install/blob/e5f169ad3b94ddbb12cec24b5fa46135bd89eb9b/openvpn-install.sh#L116
https://github.com/angristan/openvpn-install/blob/e5f169ad3b94ddbb12cec24b5fa46135bd89eb9b/openvpn-install.sh#L120-L128
The non defaults are the `root-hints` option because they are built in and so downloading them isn't necessary and the `num-threads: 2` option that defaults to 1 (see https://wiki.archlinux.org/index.php/Unbound#Issues_concerning_num-threads)

randshell:
> https://wiki.archlinux.org/index.php/Unbound#Issues_concerning_num-threads

`num-threads` is `4` on Centos 7. In this case we can keep the default value that comes with the package on each distro or maybe overwrite it to `2` if it defaults to `1`.

As a note, we specify `qname-minimisation: yes` for privacy reasons but it seems that it falls back to `no` sending all QNAME if the nameservers are broken and don't accept the minimsation. See https://www.nlnetlabs.nl/documentation/unbound/unbound.conf/#qname-minimisation

randshell (changes requested):
Can you update the path to our config in the questions please?
https://github.com/angristan/openvpn-install/blob/e5f169ad3b94ddbb12cec24b5fa46135bd89eb9b/openvpn-install.sh#L270

randshell (approved):
@HenryNe Looks good.
Now I'm waiting for feedback from @angristan to resolve my previous comments.

HenryNe:
`username: unbound` is default. Testet on Arch Linux:
1. Removed the "username: ..." from /etc/unbound/unbound.conf
2. `systemctl restart unbound`
3. Get Current UID `ps l -p `pidof unbound``
F   UID     PID    PPID PRI  NI    VSZ   RSS WCHAN  STAT TTY        TIME COMMAND
4   **976**    1569       1  20   0  92292 15012 -      Ssl  ?          0:00 /usr/bin/unbound -d -p
4. Check user ID `id 976`
uid=976(**unbound**) gid=976(unbound) groups=976(unbound)

randshell:
@HenryNe these are defaults too https://github.com/angristan/openvpn-install/pull/604#issuecomment-607532656.

You can use [unbound.conf.txt](https://github.com/angristan/openvpn-install/files/4417411/unbound.conf.txt) to see them

HenryNe:
"directory: /etc/unbound" is default on Arch Linux.

1. Removed the "directory: ..." from /etc/unbound/unbound.conf
2. `systemctl restart unbound`
3. Get Current Working Directory from running daemon:

```
ls -l /proc/`pidof unbound`/cwd
lrwxrwxrwx 1 root root 0 Apr  4 19:24 /proc/1569/cwd -> /etc/unbound
```

randshell:
> `ls -l /proc/`pidof unbound`/cwd`

Or you can use `unbound-checkconf -o directory`. See https://www.nlnetlabs.nl/documentation/unbound/unbound-checkconf/

HenryNe:
unbound is started with `/usr/bin/unbound -d -p` under Arch Linux.
So the `do-daemonize: no` not need.ed.

HenryNe:
The only last would be `num-threads: 2`, which is 1 default.

HenryNe:
I changed the PR into ready for pull now, because this change uses the same options as before. Only the include and config file was movoed.

PR #608, opened 2020-04-02
Improve sed line deletion

randshell:
https://github.com/angristan/openvpn-install/blob/e123635e7ca50792523ea034bf2ed19749aeb690/openvpn-install.sh#L1131
There is this line too but it's fixed in #604

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -1128,7 +1128,7 @@ function revokeClient () {
 
 function removeUnbound () {
 	# Remove OpenVPN-related config
-	sed -i 's|include: \/etc\/unbound\/openvpn.conf||' /etc/unbound/unbound.conf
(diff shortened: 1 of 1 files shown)


HenryNe (approved):

HenryNe:
> 
> There is this line too but it's fixed in #604

Please change the other line also.
My PR is in draft.

HenryNe (approved):

angristan:
As usual, please provide justification for your PRs, I cannot guess every time... What do you mean by "fix"? Is it broken? Is it just a better syntax? 

Thanks

randshell:
First is better syntax, second this change deletes the line instead of replacing it with nothing. The difference is that it doesn't leave an empty line between the others lines.

Before:
```
client1

client3
```
After:
```
client1
client3
```

angristan:
Great 👍

PR #620, opened 2020-04-16
Bash good practices

angristan:


angristan:
Todo: remove `|| return`

HenryNe:
The `set -euo pipefail` is a wery good idea.
But, I would not change the top line. Why do you do it?

PS: As reference for your _bash good practices_, I found this post:
https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/ 
Without the 'x' for us here.

angristan:
The use of `/usr/bin/env` is generally considered to make scripts more portable

HenryNe:
Ok, understand. It is for systems where bash is under other path, for example `/usr/bin/bash`.

HenryNe:
Have you generaly rules for bash style?. I see variables in quotation marks, sometime without. Constants somtimes with double quotation marks, single, and without.

https://github.com/angristan/openvpn-install/blob/5e2e67f78df9f92a8d7ce92338a8f6e653c0c149/openvpn-install.sh#L7

https://github.com/angristan/openvpn-install/blob/5e2e67f78df9f92a8d7ce92338a8f6e653c0c149/openvpn-install.sh#L40

https://github.com/angristan/openvpn-install/blob/5e2e67f78df9f92a8d7ce92338a8f6e653c0c149/openvpn-install.sh#L201

https://github.com/angristan/openvpn-install/blob/5e2e67f78df9f92a8d7ce92338a8f6e653c0c149/openvpn-install.sh#L319

https://github.com/angristan/openvpn-install/blob/5e2e67f78df9f92a8d7ce92338a8f6e653c0c149/openvpn-install.sh#L335

angristan:
Yes, everything is here: https://github.com/angristan/openvpn-install/blob/master/.github/workflows/lint.yml

angristan:
Should be covered by https://github.com/angristan/openvpn-install/pull/1369 now

PR #628, opened 2020-04-22
Ignore block-outside-dns in Linux

### LINKED ISSUE #254: Activation of openvpn seems to fail - Error block-outside-dns
I load openvpn after having completed openvpn-install.sh. I do that in PuTTY and PuTTy stops after mentioning Rout_Gateway and my IP Adress. After that I cannot connect to the server with PuTTY anymore. The first line "... block-outside-dns" reads like an error. Sorry I have no clue how to proceed.
Thanks for some help.

image

randshell:
Fix #254

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -998,6 +998,7 @@ cipher $CIPHER
 tls-client
 tls-version-min 1.2
 tls-cipher $CC_CIPHER
+ignore-unknown-option block-outside-dns
 setenv opt block-outside-dns # Prevent Windows 10 DNS leak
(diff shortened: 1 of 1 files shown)


angristan:
Would that ignore the option only if it's unsuported?

randshell:
> –ignore-unknown-option opt1 opt2 opt3 … optN
When one of options opt1 … optN is encountered in the configuration file the configuration file parsing does not fail if this OpenVPN version does not support the option.

Yes. It's a Windows option so this is to ignore it in Linux.

angristan (approved):

PR #629, opened 2020-04-22
Add FAQ for router clients

### LINKED ISSUE #508: FAQ clarification for TLS Crypt
This is not an issue but a problem of ASUS routers which I spent a lot of time to understand and nice to be reflected in FAQ. 

Asus Routers do not support Tls-crypt. 
In such case they throw error:
Options error: No closing quotation (") in config.ovpn:46

To fix it is need to go to custom encryption setting during setup and choose TLS Auth.

randshell:
Fix #508

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -91,10 +91,13 @@ Sysctl options are at `/etc/sysctl.d/20-openvpn.conf`
 
 **Q:** My router can't connect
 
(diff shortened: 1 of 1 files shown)


angristan:
Maybe put that in a markdown list? it would be easier to read

randshell:
Indeed.

PR #630, opened 2020-04-22
Add FAQ to access remote LAN

### LINKED ISSUE #575: Adding internal network
Hello,

I'm using Ubuntu 18 LTS as server and I've added both public and private network in the Proxmox configuration. Also, I see the networks in the server.

However, the OpenVPN only allow me to surf public networks, not private network. Is there any config I need to add after using your script to allow me to add the 192.168.*.* range?

randshell:
Fix #575


TinCanTech:
Using 192.168.1.0/24 (and similar subnets) for the server LAN is a bad idea. 
I realise this is just a FAQ entry but it seems a good idea to add a warning about conflicting routes here.

randshell:
> Using 192.168.1.0/24 (and similar subnets) for the server LAN is a bad idea.

This is the default configuration if you install a VPN at your home

> add a warning about conflicting routes here

Example if both client network and LAN network of the VPN server are 192.168.1.0/24:
By default private network isn't routed through the VPN so if you connect to 192.168.1.1 you should see your home router web page. If you want to access the router of the remote network you add this routing, of course then you can't access your home router page because you redirect the IP to the VPN.

This is what this FAQ is about. What do you mean?

TinCanTech:
> By default private network isn't routed through the VPN so if you connect to 192.168.1.1 you should see your home router web page. If you want to access the router of the remote network you add this routing, of course then you can't access your home router page because you redirect the IP to the VPN.

This only make sense if both server and client LANs are 192.168.1.0/24

I understand that the FAQ is only an example but it should be made clear that the user LAN may be any other RFC1918 subnet.  Explicitly state: `push "route 192.168.1.0 255.255.255.0" ...` is only an example subnet.

randshell:
> This only make sense if both server and client LANs are 192.168.1.0/24

> Example if both client network and LAN network of the VPN server are 192.168.1.0/24:

I mentioned it.
> Explicitly state:

You are right,  thanks for reviewing.

FAQ.md
```diff
@@ -88,3 +88,9 @@ If your client is <2.3.3, remove `tls-version-min 1.2` from your `/etc/openvpn/s
 Sysctl options are at `/etc/sysctl.d/20-openvpn.conf`
 
 ---
+
+**Q:** How can I access computers on the remote nework?
```
angristan: Remote? or LAN?

randshell: LAN is more clear,  I've changed it.

angristan: I'm not sure why you call this remote ? From what I understand this allows you to connect to your LAN while being connected to the VPN

randshell: Are you saying that if both your client LAN and the LAN behind the VPN server have the same subnet `192.168.1.0` and you access the router web page, the router that answers is the remote one and you want to exclude that subnet from VPN routing so that your local router answers instead?

From the original issue I understood the opposite and so you need to add a route to access the router on the remote LAN.

From @ruicruzpt comment:
> OpenVPN only allow me to surf public networks, not private network. Is there any config I need to add after using your script to allow me to add the 192.168.. range?

angristan: For me, this isn't about "remote" LAN. OpenVPN will override the default route/gateway so all traffic will be routed to the OpenVPN server. You LAN will be inaccessible. `push "route 192.168.1.0 255.255.255.0"` will add another route on top of the new gateway so add traffic to your LAN will be forwarded to your local router and this be accessible again.

randshell: But if openvpn overrides the default gateway then routing 192.168.1.0 to 255.255.255.0 redirects to the default gateway that is from openvpn. To properly exclude an IP the command is `route 192.168.1.0 255.255.255.0 net_gateway`.

With `route 192.168.1.0 255.255.255.0` we are simply advertising that the subnet 192.168.1.0 is available and should be routed through the vpn.
I don't host a vpn at home so I can't test. These are the links I read:
- https://openvpn.net/community-resources/expanding-the-scope-of-the-vpn-to-include-additional-machines-on-either-the-client-or-server-subnet/
- https://community.openvpn.net/openvpn/wiki/RoutedLans
- https://forums.openvpn.net/viewtopic.php?t=7065

angristan: I'll take a look

angristan: Ok, I had it the other way around, I just tested. Thanks!

FAQ.md
```diff
@@ -104,3 +104,9 @@ Sysctl options are at `/etc/sysctl.d/20-openvpn.conf`
 - `Options error: Unrecognized option or missing parameter(s) in config.ovpn:36: tls-version-min (2.3.2)` :
 
   see question "Can I use an OpenVPN 2.3 client?"
+
+---
+
+**Q:** How can I access computers the OpenVPN server's remote LAN?
```
randshell, line 110: @angristan Maybe `computers in the`?

angristan: oh thanks for the catch

PR #642, opened 2020-04-27
Update openvpn-install.sh

techroy23:
added --state

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -946,11 +946,13 @@ verb 3" >>/etc/openvpn/server.conf
 
 	# Script to add rules
 	echo "#!/bin/sh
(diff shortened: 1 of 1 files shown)


angristan:
Any explanation for this change?

techroy23:
> Any explanation for this change?

to filter out packets

PSS ; sorry i forgot the entire explanation but that is the typical setup i used for my iptables rules

angristan:
Why did you remove the position (`1`) of the rules?

techroy23:
> Why did you remove the position (`1`) of the rules?

i forgot to add 1 sorry

randshell:
@angristan as a note, Nyr does the same thing

@techroy23 why you changed `-I` to `-A`?

PR #644, opened 2020-04-27
Update README.md

techroy23:
inserted bash script for adding multiple users


angristan (changes requested):
That would be better in FAQ.md

techroy23 (commented):
ok ill move it to FAQ

angristan:
That would be better in FAQ.md

PR #653, opened 2020-05-02
Migrate to openvpn 2.4 paths and systemd services

### LINKED ISSUE #605: Use `/etc/openvpn/server` instead of `/etc/openvpn`
Some distros like Centos and Fedora use a different directory for server and client configuration. It is cleaner if we do the same for all distros.

randshell:
Openvpn 2.4 uses by default two directories `client` and `server` to separate the openvpn configuration. We now move to this behavior and patch distros with openvpn 2.3 to use the new path. Only Ubuntu 16.04 and Debian 8 are affected.
Fixes #605 

@angristan for how long do you want to support 2.3? Nyr dropped support for it.
I know from the issues that it's used on routers but I'm not sure about the security implications.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -63,7 +63,7 @@ down /usr/share/openvpn/contrib/pull-resolv-conf/client.down
 - AES CBC
 - tls-auth
 
-If your client is <2.3.3, remove `tls-version-min 1.2` from your `/etc/openvpn/server.conf` and `.ovpn` files.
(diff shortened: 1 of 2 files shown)


randshell:
> Only Ubuntu 16.04 and Debian 8 are affected.

I read the code again and these two versions use the repository from openvpn.net that provides the latest version of openpvn. This means we can drop all 2.3 paths and the related systemd services.

angristan:
If all distributions are packaged properly now, I agree we should do these changes

PR #655, opened 2020-05-02
Delete old easy-rsa remove

randshell:
It isn't packaged anymore with openvpn in the supported distros


randomshell:
We also change the path of our `easy-rsa` in #653

angristan (approved):
let's do it

Issue #563, opened 2020-03-15
[Request] Have an option in the script (once OpenVPN is installed) to list users (clients)


AMDBartek:
**Example:**

What do you want to do?
   1) Add a new user
   2) Revoke existing user
   3) List users
   4) Remove OpenVPN
   5) Exit
Select an option [1-5]:

randshell:
PR #587

PR #669, opened 2020-05-24
Add Certificate List Function V2

psgoundar:
Added  a Certificate List function with requested Changed.
Function will List issues Certificates with CN and Validity Dates.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -1296,7 +1296,9 @@ if [ ! -f "${INDEX}" ]; then
         exit 1
 fi
 printf "\\e[1m::: Certificate Status List :::\\e[0m\\n"
-printf "\\e[4mStatus\\e[0m ::  \\e[4mName\\e[0m\\e[0m        ::  \\e[4mExpiration \\e[0m\\n" 
+printf "\\n%6s\\t%14s\\t%22s\\n" "Status" "Name" "Expiration"
+printf '=%.0s' {1..50}
+printf '\n'
 while read -r line || [ -n "$line" ]; do
     STATUS=$(echo "$line" | awk '{print $1}')
     NAME=$(echo "$line" | awk '{print $5}' | awk -FCN= '{print $2}')
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -1291,6 +1291,34 @@ function removeOpenVPN() {
 	fi
 }
 
+function listcerts () {
+# Original Script from PiVPN: list clients script
+# Modified Script to add Certificate expiration Date -- psgoundar  
+INDEX="/etc/openvpn/easy-rsa/pki/index.txt"
+printf "\\n"
+if [ ! -f "${INDEX}" ]; then
+        echo "The file: $INDEX was not found!"
+        exit 1
+fi
+printf "\\e[1m::: Certificate Status List :::\\e[0m\\n"
+printf "\\e[4mStatus\\e[0m ::  \\e[4mName\\e[0m\\e[0m        ::  \\e[4mExpiration \\e[0m\\n" 
+while read -r line || [ -n "$line" ]; do
+    STATUS=$(echo "$line" | awk '{print $1}')
+    NAME=$(echo "$line" | awk '{print $5}' | awk -FCN= '{print $2}')
+    EXPD=$(echo "$line" | awk '{if (length($2) == 15) print $2; else print "20"$2}' | cut -b 1-8 | date +"%b %d %Y" -f -)
+        
+    if [ "${STATUS}" == "V" ]; then
+         printf "     Valid   ::   %s :: %s\\n" "$NAME" "$EXPD"
+    elif [ "${STATUS}" == "R" ]; then
+        #printf "     Revoked   ::   %s :: %s\\n" "$NAME" "$EXPD"
+		continue
+    else
+        printf "     Unknown   ::   %s :: %s\\n" "$NAME" "$EXPD"
+    fi
+done <${INDEX} | column -t
+printf "\\n"
+}
+
 function manageMenu() {
 	echo "Welcome to OpenVPN-install!"
 	echo "The git repository is available at: https://github.com/angristan/openvpn-install"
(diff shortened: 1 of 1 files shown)


randshell:
Fixes #563

openvpn-install.sh
```diff
@@ -1309,9 +1338,12 @@ function manageMenu() {
 		revokeClient
 		;;
 	3)
-		removeOpenVPN
+			listcerts
```
randshell: Please use the same casing as the other functions and remove the added space

openvpn-install.sh
```diff
@@ -1295,10 +1323,11 @@ function manageMenu() {
 	echo "What do you want to do?"
 	echo "   1) Add a new user"
 	echo "   2) Revoke existing user"
-	echo "   3) Remove OpenVPN"
-	echo "   4) Exit"
-	until [[ $MENU_OPTION =~ ^[1-4]$ ]]; do
-		read -rp "Select an option [1-4]: " MENU_OPTION
+	echo "   3) List Current Issued Certificates"
```
randshell: Please use the same casing as the other options

openvpn-install.sh
```diff
@@ -1286,6 +1286,34 @@ function removeOpenVPN() {
...
+    EXPD=$(echo "$line" | awk '{if (length($2) == 15) print $2; else print "20"$2}' | cut -b 1-8 | date +"%b %d %Y" -f -)
+        
+    if [ "${STATUS}" == "V" ]; then
+         printf "     Valid   ::   %s :: %s\\n" "$NAME" "$EXPD"
+    elif [ "${STATUS}" == "R" ]; then
+        #printf "     Revoked   ::   %s :: %s\\n" "$NAME" "$EXPD"
+		continue
+    else
+        printf "     Unknown   ::   %s :: %s\\n" "$NAME" "$EXPD"
+    fi
```
randshell, line 1313: Can we use `\t` instead of multiple spaces?

randshell (changes requested):
There are some issues with the PR. Other than what I wrote in the comments, can you remove the `openvpn-install` submodule?

If you can, merge the commits with the changes in one after.

Thanks for contributing!

openvpn-install.sh
psgoundar: Changed to Tabs

openvpn-install.sh
psgoundar: Changed Casing

openvpn-install.sh
psgoundar: Changed Casing

psgoundar (commented):
Changes Complete

openvpn-install.sh
```diff
@@ -1286,6 +1286,33 @@ function removeOpenVPN() {
...
+# Original Script from PiVPN: list clients script
+# Modified Script to add Certificate expiration Date -- psgoundar  
+INDEX="/etc/openvpn/easy-rsa/pki/index.txt"
+printf "\\n"
+if [ ! -f "${INDEX}" ]; then
+        echo "The file: $INDEX was not found!"
+        exit 1
+fi
+printf "\\e[1m::: Certificate Status List :::\\e[0m\\n"
+printf "\\e[4mStatus\\e[0m ::  \\e[4mName\\e[0m\\e[0m        ::  \\e[4mExpiration \\e[0m\\n"
```
```suggestion
printf "\\n%6s\\t%14s\\t%22s\\n" "Status" "Name" "Expiration"
printf '=%.0s' {1..50}
printf '\n'
```
I tried to align them more. What do you think?

openvpn-install.sh
```diff
@@ -1286,6 +1286,33 @@ function removeOpenVPN() {
...
+printf "\\n"
+if [ ! -f "${INDEX}" ]; then
+        echo "The file: $INDEX was not found!"
+        exit 1
+fi
+printf "\\e[1m::: Certificate Status List :::\\e[0m\\n"
+printf "\\e[4mStatus\\e[0m ::  \\e[4mName\\e[0m\\e[0m        ::  \\e[4mExpiration \\e[0m\\n" 
+while read -r line || [ -n "$line" ]; do
+    STATUS=$(echo "$line" | awk '{print $1}')
+    NAME=$(echo "$line" | awk '{print $5}' | awk -FCN= '{print $2}')
```
```suggestion
    NAME=$(echo "$line" | cut -d '=' -f2)
```
This doesn't work when clients are revoked, I changed it.

openvpn-install.sh
```diff
@@ -1286,6 +1286,33 @@ function removeOpenVPN() {
...
+    EXPD=$(echo "$line" | awk '{if (length($2) == 15) print $2; else print "20"$2}' | cut -b 1-8 | date +"%b %d %Y" -f -)
```
randshell: The expiration date is wrong for revoked clients, we should use the column provided instead of calculating it.

randshell (changes requested):

psgoundar (commented):
I"m fine with the proposed changes. Thanks for doing that.

openvpn-install.sh
```diff
@@ -1286,6 +1286,35 @@ function removeOpenVPN() {
...
+        exit 1
+fi
+printf "\\e[1m::: Certificate Status List :::\\e[0m\\n"
+printf "\\n%6s\\t%14s\\t%22s\\n" "Status" "Name" "Expiration"
+printf '=%.0s' {1..50}
+printf '\n'
+while read -r line || [ -n "$line" ]; do
+    STATUS=$(echo "$line" | awk '{print $1}')
+    NAME=$(echo "$line" | cut -d '=' -f2)
+    EXPD=$(echo "$line" | awk '{if (length($2) == 15) print $2; else print "20"$2}' | cut -b 1-8 | date +"%b %d %Y" -f -)
```
```suggestion
    EXPD=$(echo "$line" | awk '{if (length($3) == 13) print "20"$3; else print "20"$2}' | cut -b 1-8 | date +"%b %d %Y" -f -)
```
You just missed to check the third column first. Now it should work for revoked certs too.

randshell (approved):
Thanks!

PR #670, opened 2020-05-26
Update openvpn-install.sh

MichaelRoberts98:
Updated openvpn-install.sh to use ip.me since ifconfig.co is not working

Curling Ifconfig.co returns a Heroku application error response instead of the correct IP.
image

I updated the ipv4 section to use the tool https://ip.me, however do not have a replacement for the ipv6 portion.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -618,7 +618,7 @@ function installOpenVPN() {
 		if [[ $IPV6_SUPPORT == "y" ]]; then
 			PUBLIC_IP=$(curl https://ifconfig.co)
 		else
-			PUBLIC_IP=$(curl -4 https://ifconfig.co)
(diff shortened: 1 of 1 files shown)


angristan:
https://ifconfig.co seems to be back online!

MichaelRoberts98:
That's good @angristan! Would we be able to add in a check on the output from ifconfig.co and if it doesn't respond with an IP then it performs the curl -4 on another source?

randshell:
> That's good @angristan! Would we be able to add in a check on the output from ifconfig.co and if it doesn't respond with an IP then it performs the curl -4 on another source?

This is a good suggestion, I also had this issue in the past.

Can we retry requesting ifconfig.co after a delay instead of using another source? It should work after a few times and it can be done with curl arguments. On our end we can add a check to make sure the answer we receive it's a valid IP, with grep for example.

PR #683, opened 2020-06-26
ci: use super-linter

angristan:


randshell:
Super-linter uses shellcheck so you can remove the previous action.

You can specify the language `Shell` to run only shellcheck.

angristan:
Let's see how this goes

Issue #633, opened 2020-04-22
Suggest browserleaks.com for dns leak tests


randshell:
The FAQ suggests https://dnsleaktest.com/ and https://ipleak.net/ to test dns leaks. Do they support IPv6 leaks? I don't have an IPv6 connection to test.

An alternative is https://browserleaks.com/dns that explicitly tests IPv6:
> This test attempts to resolve 100 randomly generated domain names asynchronously, 50 with A record (IPv4-only) and 50 with both A and AAAA records (IPv4+IPv6).

angristan:
https://ipleak.net/ does support IPv6, but https://browserleaks.com/dns seems like a good alternative too

PR #685, opened 2020-06-27
Add new dns leak tests to FAQ

robiiinos:
Suggested from #633 

Should we specify which of them support IPv6 ?

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -10,7 +10,7 @@ You can, of course, it's even recommended, update the `openvpn` package with you
 
 **Q:** How do I check for DNS leaks?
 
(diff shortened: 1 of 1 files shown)


angristan:
All of them support IPv6!

angristan:
Sorry, https://dnsleaktest.com/ doesn't.

angristan:
I suppose we should remove `dnsleaktest.com` since it does not support IPv6 and does not provide anything more than the two other alternatives

robiiinos:
Done ! Also specified that both perform IPv4 and IPv6 check, just in case.

angristan (approved):
Thanks for the contribution 👍

PR #699, opened 2020-07-17
Updated client name input restrictions and hint

Serpentiel:
The hint appears not to describe the actual regular expression used, so I have changed it to do so.

Also, I've made changes to the regular expression itself as having dash in name of client doesn't break things up.


randshell:
Great! Thanks for contributing @Serpentiel!

angristan (approved):
Thanks!

randshell:
@Serpentiel since this change has been accepted, would you like to open a PR for [wireguard-install](https://github.com/angristan/wireguard-install) too?

Serpentiel:
@randomshell yes sure, will do

PR #702, opened 2020-07-25
Fix GitHub Actions

robiiinos:
- Pin GitHub Actions checkout version as using ['master' will not be the default branch](https://github.com/angristan/openvpn-install/actions/runs/182285350) anymore.

Screenshot 2020-07-25 at 15 31 15

- Avoid breaking changes by pinning a released version of the Actions for ssh-action.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,9 @@ jobs:
           - centos-7-x64
           - centos-8-x64
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
+      with:
+        ref: master
 
     - name: Setup doctl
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -21,7 +21,9 @@ jobs:
           - centos-7-x64
           - centos-8-x64
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
+      with:
+        ref: master
```
angristan: I think the warning was about this `@master` specifically, `with.ref: master` is something else isn't it?

robiiinos: Totally misunderstood the problem there, fixed.

PR #708, opened 2020-08-03
Add retries to curl ifconfig.co

angristan:
Fix for #670


randshell:
Can you edit ifconfig.co to use HTTPS in the README too?

Maybe we can add a regex to the response to make sure we get a valid IP?

PR #714, opened 2020-08-13
Update openvpn-install.sh

hadish777:


angristan:
Hi, can you provide an explanation for this change? I'm not sure I follow. Thanks

randshell:
As far as I know, easy-rsa didn't change the naming of the variables.

hadish787:
Hi there. I just needed to change the name of server in the script.

ufukty:
Line 711 could break the functionality of EasyRSA, if there is no change in variable names. And as I know, `CN` stands for `Common Name` not `Connection` (if `CON` refers to that.)

Issue #720, opened 2020-08-25
A fix about a problem with apt and dpkg lock


ufukty:
**Checklist**

- [x] I read the [README](https://github.com/angristan/openvpn-install/blob/master/README.md)
- [x] I read the [FAQ](https://github.com/angristan/openvpn-install/blob/master/FAQ.md)
- [x] I searched the [issues](https://github.com/angristan/openvpn-install/issues?q=is%3Aissue+)
- [x] My issue is about the script, and not OpenVPN itself

**Problem**

When Ubuntu starts, it triggers the apt for checking latest security patches. Since apt has no mechanism to keep sequent calls on-hold, they are ending up with error. Script can not install required packages and tries to continue without them. 

This issue more likely belongs to `apt-get` than this repo, but seems like there are issues reported about apt at bugs.debian.org [^1] [^2] none of them resolves the `apt-get`'s problem.

**Fix**

Fast fix could be wrapping apt calls with `until` and `sleep`:
```sh
until apt-get install openvpn; do sleep 1; done
```
I will make a pull request to resolve issue.

**Symptoms**

1. VPN connection can not be established.

1. `<tls-crypt>` section of `.ovpn` file remains empty

1. Packages may not be installed at all (`openvpn`, `easy-rsa` etc.) Related error for that:
    ```
    E: Could not get lock /var/lib/apt/lists/lock. It is held by process 1502 (apt-get)
    N: Be aware that removing the lock file is not a solution and may break your system.
    E: Unable to lock directory /var/lib/apt/lists/
    ```

1. Error messages caused by commands at the [line 931](https://github.com/angristan/openvpn-install/blob/423419de74ad945b905c46d097defaf490f60636/openvpn-install.sh#L931)  
    ```
    cp: cannot stat '/lib/systemd/system/openvpn@.service': No such file or directory
    sed: can't read /etc/systemd/system/openvpn@.service: No such file or directory
    sed: can't read /etc/systemd/system/openvpn@.service: No such file or directory
    Failed to enable unit: Unit file openvpn@server.service does not exist.
    Failed to restart openvpn@server.service: Unit openvpn@server.service not found.
    ```

**Server**
Ubuntu 20.04.1 LTS (GNU/Linux 5.4.0-42-generic x86_64)

**Client**
Mac — Viscosity
iOS — Official OpenVPN client

**References**
[^1]: Bug report: (apt: support option to wait for lockfile)  
[...]

PR #721, opened 2020-08-25
Fix a bug with dpkg lock prevents apt commands run

ufukty:
Related with the issue https://github.com/angristan/openvpn-install/issues/720

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -99,8 +99,7 @@ function installUnbound() {
 	if [[ ! -e /etc/unbound/unbound.conf ]]; then
 
 		if [[ $OS =~ (debian|ubuntu) ]]; then
-			apt-get install -y unbound
-
+			until apt-get install -y unbound; do sleep 1; done
 			# Configuration
 			echo 'interface: 10.8.0.1
 access-control: 10.8.0.1/24 allow
@@ -648,16 +647,16 @@ function installOpenVPN() {
 	# the first time.
 	if [[ ! -e /etc/openvpn/server.conf ]]; then
 		if [[ $OS =~ (debian|ubuntu) ]]; then
-			apt-get update
-			apt-get -y install ca-certificates gnupg
+			until apt-get update; do sleep 1; done
+			until apt-get -y install ca-certificates gnupg; do sleep 1; done
 			# We add the OpenVPN repo to get the latest version.
 			if [[ $VERSION_ID == "16.04" ]]; then
 				echo "deb http://build.openvpn.net/debian/openvpn/stable xenial main" >/etc/apt/sources.list.d/openvpn.list
 				wget -O - https://swupdate.openvpn.net/repos/repo-public.gpg | apt-key add -
-				apt-get update
+				until apt-get update; do sleep 1; done
 			fi
 			# Ubuntu > 16.04 and Debian > 8 have OpenVPN >= 2.4 without the need of a third party repository.
-			apt-get install -y openvpn iptables openssl wget ca-certificates curl
+			until apt-get install -y openvpn iptables openssl wget ca-certificates curl; do sleep 1; done
 		elif [[ $OS == 'centos' ]]; then
 			yum install -y epel-release
 			yum install -y openvpn iptables openssl wget ca-certificates curl tar 'policycoreutils-python*'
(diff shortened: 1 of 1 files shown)


angristan:
Maybe `apt ... || exit 1` would be better ? APT could fail for a number of reasons

ufukty:
This problem occurs nearly every time when executing the script in a machine just launched. And this makes harder to use this script with Terraform and other orchestration/provisioning tools. Because those tools execute tasks immediately. Although you can say it’s those tools’ job to make sure the machine is ready for using apt before launching script.

angristan:
As said earlier exiting the script on APT fail would be great, but going into an infinite loop on APT fail doesn't seem like a good design to me.

I've never encountered the issue, so this has to do with your provider, and more specifically the image you're using. I would suggest using a custom, up-to-date image that doesn't auto-update during the first boot. This has to be handled on your side, not by the script. 🙂

randshell:
@angristan as @hellresistor said https://github.com/angristan/openvpn-install/issues/720#issuecomment-683961800, it happens by default on Ubuntu and possibly on Debian if `unattanded-upgrades` is configured to run at that time.

nevesnunes:
> As said earlier exiting the script on APT fail would be great, but going into an infinite loop on APT fail doesn't seem like a good design to me.
> 
> I've never encountered the issue, so this has to do with your provider, and more specifically the image you're using. I would suggest using a custom, up-to-date image that doesn't auto-update during the first boot. This has to be handled on your side, not by the script. slightly_smiling_face

I was also hit with this issue and see the benefits of having the script fail early, so that the root cause is more evident to the user: e.g. `/etc/openvpn/tls-crypt.key: No such file or directory` does not directly point to issues installing `openvpn`.

We can think of alternatives for the infinite loop:

1. Loop with a fixed number of max attempts, e.g.
```bash
attempts=10
while [ $attempts -gt 0 ]; do
  apt-get install -y openvpn && break
  attempts=$((attempts - 1))
done
```
2. Run the command once, and do some post-install validations, e.g. `command -v openvpn`.

ufukty:
This is not about the OS image. All linux dists have cron tasks that may run any time and sometimes they can conflict with each other. There is undeniable possibility that apt is already running before user fire up the script, even if the system is up-to-date, even it is a custom image or the default one.

The problem is, in the case of failure happened, inexperienced users may not understand why it is not working. A notice about the problem should displayed to user. 

Solution of @nevesnunes is good although it lacks the 1 second of sleep which may do the magic. And since it is 5 lines and repeats at multiple places of script, it could be organized within a function. 

In summary:

```sh
function apt_get_with_retry {
  attempts=0
  until apt-get $@; do 
    sleep 1
    attempts=$((attempts + 1))
    if [[ $attempts -gt 10 ]]; then
      echo "apt-get is busy right now, please try again later"
      exit 1
    fi
  done
}
```

Calls

```sh
apt_get_with_retry install -y openvpn iptables openssl wget ca-certificates curl
# ...
apt_get_with_retry install -y ca-certificates gnupg
# ...
apt_get_with_retry autoremove --purge -y unbound
```

I didn't check if it's working tho. 😎

randshell:
> There is undeniable possibility that apt is already running before user fire up the script, even if the system is up-to-date, even it is a custom image or the default one.

Looks like some real examples are `apt-daily.service` and `apt-daily-upgrade.service`. I wonder why they are not better configured to maybe run at night or at a custom time like `unattanded-upgrades`.

`apt-daily.service` in particular runs at random times
```
OnCalendar=*-*-* 6,18:00
RandomizedDelaySec=12h
```
but `apt-daily-upgrade.service` only between 6 and 7 am
```
OnCalendar=*-*-* 6:00
RandomizedDelaySec=60m
```
@ufukty when did you experienced this issue, not long after reboot or during the day? I think on Ubuntu it's known to wait for some time after boot for the lock file.
> The problem is, in the case of failure happened, inexperienced users may not understand why it is not working. 

Funny enough, the automatic upgrade on Debian based distros is made for inexperienced users. I don't remember anything similar on others.

ufukty:
@randomshell That was right after reboot and recurred for multiple times. I don't think there is significant waiting between booting the system and running the script because of automation. 

Seems like they rolled a new version of apt which introduces a `wait for lock` feature on 7 March 2020. But they only mention `apt` not `apt-get`.

```
* apt(8): Wait for lock (Closes: #754103)
```
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=754103
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=850172

randshell:
I like more the idea of using `apt`, it's the recommended command since Ubuntu 16.04 too.

PR #727, opened 2020-09-08
FAQ - Added how to set up part of the routing

rootux:
# Implements
Updated `FAQ.md` - Added how to set only some of the traffic to go through the VPN and the rest using the normal connection

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -124,6 +124,13 @@ for i in ${userlist[@]};do
     MENU_OPTION=1 CLIENT=$i PASS=1 ./openvpn-install.sh
 done
 ```
+
+---
+
+**Q:** How do I change the default `.ovpn` file created for future clients?
+
+**A:** You can edit the template out of which `.ovpn` files are created by editing `/etc/openvpn/client-template.txt`
+
 ---
 
 **Q:** For my clients - I want to set my internal network to pass through the VPN and the rest to go through my internet?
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -124,3 +124,14 @@ for i in ${userlist[@]};do
     MENU_OPTION=1 CLIENT=$i PASS=1 ./openvpn-install.sh
 done
 ```
+---
+
+**Q:** For my clients - I want to set my internal network to pass through the VPN and the rest to go through my internet?
+
+**A:** You would need to edit the `.ovpn` file. You can edit the template out of which those files are created by editing `/etc/openvpn/client-template.txt` file and adding
```
randshell, line 138: In my opinion `client-template.txt` deserves its own question. @angristan

rootux:
I agree. Shall I add it?

FAQ.md
angristan: Agreed!

rootux: I agree. Shall I add them?

FAQ.md
```diff
@@ -124,3 +124,21 @@ for i in ${userlist[@]};do
     MENU_OPTION=1 CLIENT=$i PASS=1 ./openvpn-install.sh
 done
 ```
+
+---
+
+**Q:** How do I change the default `.ovpn` file created for future clients?
```
rootux, line 130: @randomshell @angristan  added general instructions for the `client-template.txt` file

angristan (approved):

PR #729, opened 2020-09-10
Adding VPN_NETWORK option and other details

joaduo:
I added some options that I believe may be wanted by others:

`$VPN_NETWORK` setting the /24 VPN network prefix eg: `10.8.0`
No DNS option (number 14 right now). When you don't want to route DNS requests through the VPN (split tunneling)
`$CLIENT_TEMPLATE_APPEND` when you want to add more configuration to the client template file

You can check diff here
https://github.com/angristan/openvpn-install/compare/master...joaduo:joaduo_2020_09_upstream_pull_request


openvpn-install.sh
```diff
@@ -214,9 +214,16 @@ function installQuestions() {
 	echo "I need to ask you a few questions before starting the setup."
 	echo "You can leave the default options and just press enter if you are ok with them."
 	echo ""
+
+	echo "Please provide the VPN network prefix. If you are going to use  10.8.0.0/24, then type prefix 10.8.0"
```
randshell, line 218: I don't see why this question should be the first one to be asked.

openvpn-install.sh
```diff
@@ -214,9 +214,16 @@ function installQuestions() {
...
+	until [[ $VPN_NETWORK =~ ^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){2}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$ ]]; do
+		echo "Format bust be like XXX.XXX.XXX"
+		VPN_NETWORK=${VPN_NETWORK:-10.8.0}
```
randshell: What is this line for? You can write `10.8.0` in the `read -i` command as it is.

Can you move it to `installOpenVPN()` instead, it's needed for the auto-install.

openvpn-install.sh
```diff
@@ -836,8 +844,13 @@ ifconfig-pool-persist ipp.txt" >>/etc/openvpn/server.conf
...
 		fi
 		;;
+	14) # No DNS
+		echo "No DNS push config"
+		;;
 	esac
-	echo 'push "redirect-gateway def1 bypass-dhcp"' >>/etc/openvpn/server.conf
+	if [[ $DNS != 14 ]]; then
+		echo 'push "redirect-gateway def1 bypass-dhcp"' >>/etc/openvpn/server.conf
+	fi
```
randshell, line 853: You are using the `None` option to connect to your devices without redirecting all the traffic. It's not really the script's goal, so maybe a FAQ would be better? @angristan

randshell (commented):
There is already a PR opened for configuring the VPN mask #550, I don't see much difference between the two.

openvpn-install.sh
joaduo: In my case I want a secure way to access to my servers, it's not my idea to redirect all traffic.
So I assume not wanting a DNS means also not wanting to redirect all traffic. (I believe users can avoid redirecting traffic anyway, but sounds natural to me to also enforce it at server level)
I use `$CLIENT_TEMPLATE_APPEND` option to add specific routes. eg:
```
CLIENT_TEMPLATE_APPEND="
route 172.34.0.0 255.255.0.0
"
```
I will clarify the option. (or perhaps i can split in two options (redirecting and DNS))

openvpn-install.sh
joaduo: This is a bug. moving to `installOpenVPN()` as suggested. thanks

openvpn-install.sh
joaduo: Thanks for your comments.
Any order is ok. Which order do you consider proper?
I just need a way to setup the IP Range, if PR #550 allows this, then fair enough.

openvpn-install.sh
randshell: > In my case I want a secure way to access to my servers, it's not my idea to redirect all traffic.

Again, not really the script's goal but I understand that need, I used to do the same thing.

> So I assume not wanting a DNS means also not wanting to redirect all traffic.

Yes you assume correctly, otherwise it would be a DNS leak.

angristan:
Dupe of https://github.com/angristan/openvpn-install/pull/550

PR #735, opened 2020-10-02
Automatic public IP discovery for servers behind NAT

phonicmouse:
I added IP discovery using [ipify](https://www.ipify.org/)

It was useful for my use case and i thought it might speed up things for other users as well.


angristan (approved):

angristan:
Thanks :)

PR #748, opened 2020-10-19
Remove key-direction from tls-crypt option

randshell:
In contrast to --tls-auth, --tls-crypt does *not* require the user to set --key-direction. Thus syntax is `--tls-crypt keyfile`

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -861,7 +861,7 @@ push "redirect-gateway ipv6"' >>/etc/openvpn/server.conf
 
 	case $TLS_SIG in
 	1)
-		echo "tls-crypt tls-crypt.key 0" >>/etc/openvpn/server.conf
+		echo "tls-crypt tls-crypt.key" >>/etc/openvpn/server.conf
 		;;
 	2)
 		echo "tls-auth tls-auth.key 0" >>/etc/openvpn/server.conf


angristan:
What happens when key direction is set? Is it simply ignored?

angristan:
(unrelated: could you rebase on master? I want to know if CI run on PRs)

TinCanTech:
Testing shows that the `key-direction` field is ignored if appended to the `--tls-crypt`config directive.  `--key-direction` is also ignored if `tls-crypt` is specified. (I looked in the source for a comment but found none)

angristan:
Thanks @TinCanTech

randshell:
> (unrelated: could you rebase on master? I want to know if CI run on PRs)

Rebase? I use to change last commit sha and force push to trigger CI, might be a bad habit though.

Anyway, it looks like actions run twice.

angristan:
> Rebase? I use to change last commit sha and force push to trigger CI, might be a bad habit though.

Yeah sure, It's because I pushed some changes to the CI triggers on master, so I wanted this branch to have the commit. 

It's indeed running twice, still better than not running at all 😄

angristan (approved):

randshell:
@angristan Looks like `pull_request_target` and `pull_request` are the same thing except for secrets access. You may want to use one of them only.

> In order to protect public repositories for malicious users we run all pull request workflows raised from repository forks with a read-only token and no access to secrets. This makes common workflows like labeling or commenting on pull requests very difficult.
>
> In order to solve this, we’ve added a new `pull_request_target` event, which behaves in an almost identical way to the `pull_request` event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request. This event can be used in combination with the private repository settings as well.

Source: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

PR #750, opened 2020-10-20
increase sysctl conf file to 99- prefix

### LINKED ISSUE #744: Heads up -- GCP disables ip_forward=1
I just hit a gotcha which I thought might be helpful to others. While your script enables ip_forward [here](https://github.com/angristan/openvpn-install/blob/2950fd445752fd0515cdbf37b28ab8a494bc8299/openvpn-install.sh#L891), instances running on GCP turn around and _disable_ that in [60-gce-network-security.conf](https://github.com/GoogleCloudPlatform/guest-configs/blob/master/src/etc/sysctl.d/60-gce-network-security.conf). I was head-scratching why my setup suddenly stopped working, and that's all it was.

Idk, what would be the implications of being more aggressive with the number prefix here? I'm going to `99-` it in my instance just because need it to work right meow, but in general?

deargle:
Prevents GCP cloud platform's default security policy for instances, which uses prefix 60-, from overriding ip_forward. Also future-proofs against any other such default policy.

Closes #744

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -891,9 +891,9 @@ verb 3" >>/etc/openvpn/server.conf
 	mkdir -p /var/log/openvpn
 
 	# Enable routing
-	echo 'net.ipv4.ip_forward=1' >/etc/sysctl.d/20-openvpn.conf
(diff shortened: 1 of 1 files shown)


angristan:
Thanks!

PR #762, opened 2020-11-14
Remove unnecessary cipher configuration on Fedora

randshell:
Based on the fact that OpenVPN 2.5 removes the `BF-CBC` cipher from the default configuration, I expected Fedora to update its ciphers line in the RC3 of the package too which they didn't.

Upon investigation on why Fedora hard-codes the ciphers in the service file, I found out that
> If an already configured OpenVPN v2.4 based server configuration deploys `--cipher` and/or `--ncp-ciphers`, the options in the configuration file will override command line options set before `--config`. This should not break any existing configuration. 

This means that the values of `cipher` and `ncp-ciphers` we already specify in our server config will overwrite the command line options from the service file, making the `sed` command unnecessary.

Source: https://fedoraproject.org/wiki/Changes/New_default_cipher_in_OpenVPN

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -917,10 +917,6 @@ verb 3" >>/etc/openvpn/server.conf
 		sed -i 's|LimitNPROC|#LimitNPROC|' /etc/systemd/system/openvpn-server@.service
 		# Another workaround to keep using /etc/openvpn/
 		sed -i 's|/etc/openvpn/server|/etc/openvpn|' /etc/systemd/system/openvpn-server@.service
-		# On fedora, the service hardcodes the ciphers. We want to manage the cipher ourselves, so we remove it from the service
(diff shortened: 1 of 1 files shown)


TinCanTech:
Using `/etc/systemd/system` is _meant_ for customised unit files.
You can put what-ever you chose into these unit files.

randshell:
I don't understand your context. We already customize the service file in that path but in particular using `sed` to change the ciphers in the service unit is unnecessary because they are overwritten by `server.conf` anyway.

TinCanTech:
After looking closer I see: https://github.com/angristan/openvpn-install/blob/8db952d2b72cebb2b90385b0a3980cd436664263/openvpn-install.sh#L913

I had presumed you were writing your own unit file from scratch. My mistake.

Issue #734, opened 2020-09-29
Change ovpn path


g-pane:
How can I set the default path of all the generate ovpn? (instead of /root but something like /root/config)
Thanks

randshell:
Replace `$homeDir` to `/root/config` in the script https://github.com/angristan/openvpn-install/blob/2950fd445752fd0515cdbf37b28ab8a494bc8299/openvpn-install.sh#L1083-L1090

g-pane:
Is possible to do in command line without edit the SH? Since can be updated

randshell:
The script doesn't allow customizing that path yet, but you can run `sed` to edit the line.
```bash
sed -i 's|homeDir="/root"$|homeDir="/root/config"|' openvpn-install.sh
```

PR #779, opened 2020-12-27
Added variable CERTIFICATE_DIR to change the output destination certificate path

AngeloAvv:
This pull request solved the problem reported in #734. You can export a CERTIFICATE_DIR variable in order to customize the output generation path of the ovpn file. If you don't provide a value for CERTIFICATE_DIR, the execution flow will be the same as before.

I think it could be helpful since I have a scenario where I need to generate certificates and put them in an external folder than /home or /root.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ If you want to customise your installation, you can export them or specify them 
 - `CUSTOMIZE_ENC=n`
 - `CLIENT=clientname`
 - `PASS=1`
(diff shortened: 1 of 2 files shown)


AngeloAvv:
@angristan any updates on this?

angristan:
Implemented in https://github.com/angristan/openvpn-install/pull/962

angristan:
Thanks for the PR :)

PR #857, opened 2021-06-17
ifconfig.co now includes Cloudflare HTML.

neneeen:
Workaround using icanhazip (which ironically is also hosted by CF)


neneeen:
Sorry I didn't see the other issues about ifconfig.co

PR #869, opened 2021-07-06
Add support for Rocky Linux 8

derek-j-pitman:
Since Rocky is attempting to replace CentOS as the "RHEL downstream" distro, I added it in the same place as CentOS in the checkOS function.


angristan (approved):
thanks

PR #873, opened 2021-07-26
Revoke client fixed

fabiogiorgione:
Delete old client references in easy-rsa index.txt, it's useful if you delete and recreate a user with the same name


dudz94:
> Delete old client references in easy-rsa index.txt, it's useful if you delete and recreate a user with the same name

Thanks for this PR. Very useful ! :)

TinCanTech:
Since `Easy-RSA version 3.0.6`, it is possible to create certificates with duplicate `commonName`.

angristan (approved):
Thanks!

PR #891, opened 2021-08-23
Add support for AlmaLinux 8

JLHC:
AlmaLinux OS is 1:1 binary compatible with RHEL and pre-Stream CentOS.

Since Rocky Linux support is added, Alma should be added to to the checkOS function too as otherwise installation on Alma will fail.


angristan:
Thanks, could you update the README too?

JLHC:
Just updated the README as well.

angristan (approved):

PR #906, opened 2021-09-05
changed script for own IPv4 vpn adress

ronnilix:
i changed the script, that its possible to define its own vpn ipv4 adress and network.


angristan:
Should be addressed with https://github.com/angristan/openvpn-install/commit/e9deb4b8ab2bf1e7138da26fb2306010a904e853

PR #930, opened 2021-10-16
Enable oracle-epel-release for Oracle Linux

rvva:
In Oracle Linux after installing oracle-epel-release it's necessary to enable this repository:

`yum-config-manager --enable ol8_developer_EPEL`

The error did not occur for users who had previously installed the epel repository correctly.
This commit fix that oversight.
Otherwise, the openvpn, policycoreutils-python-utils packages will not be installed.

I reproduced this bug in the Oracle-Linux-8.4-2021.10.04-0 image.


angristan (approved):
Thanks

PR #945, opened 2021-11-20
revokeClient: Do not remove revoked client record from index.txt

TinCanTech:
Deleting a revoked (^R) client record from index.txt means that the
client will not be listed in the Certificate Revocation List.  This
effectively "unrevokes" the client and allows the client to continue
using the VPN.

Signed-off-by: Richard T Bonhomme <tincantech@protonmail.com>


swimmer2910:
Have the same problem. Please fix it

PR #1001, opened 2022-05-11
Update openvpn-install.sh

adamnugraha:
Added Cisco OpenDNS

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -847,6 +847,10 @@ ifconfig-pool-persist ipp.txt" >>/etc/openvpn/server.conf
 		echo 'push "dhcp-option DNS 45.90.28.167"' >>/etc/openvpn/server.conf
 		echo 'push "dhcp-option DNS 45.90.30.167"' >>/etc/openvpn/server.conf
 		;;
+	12) # Cisco-OpenDNS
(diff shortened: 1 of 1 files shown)


openvpn-install.sh
```diff
@@ -847,6 +847,10 @@ ifconfig-pool-persist ipp.txt" >>/etc/openvpn/server.conf
 		echo 'push "dhcp-option DNS 45.90.28.167"' >>/etc/openvpn/server.conf
 		echo 'push "dhcp-option DNS 45.90.30.167"' >>/etc/openvpn/server.conf
 		;;
+	12) # Cisco-OpenDNS
+		echo 'push "dhcp-option DNS 208.67.222.222"' >>/etc/openvpn/server.conf
+		echo 'push "dhcp-option DNS 208.67.220.220"' >>/etc/openvpn/server.conf
+		;;
 	13) # Custom DNS
```
angristan, line 854: You have two entries numbered `12`, so this won't work. Also, you need to update this part: https://github.com/angristan/openvpn-install/blob/82aaa341fee62d96cc1a06fd99c5292c22453c94/openvpn-install.sh#L314-L327

angristan (changes requested):

PR #1003, opened 2022-05-25
ncp-cipher option has renamed to data-cipher

shgpavel:
At startup openvpn prints "This option was renamed in openvpn 2.5". In debian, arch, etc its 2.5+ versions of openvpn package in repos already, that's why we need to change this option. https://build.openvpn.net/man/openvpn-2.5/openvpn.8.html

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -891,7 +891,7 @@ cert $SERVER_NAME.crt
 key $SERVER_NAME.key
 auth $HMAC_ALG
 cipher $CIPHER
-ncp-ciphers $CIPHER
+data-ciphers $CIPHER
 tls-server
 tls-version-min 1.2
 tls-cipher $CC_CIPHER


TinCanTech (approved):
In a purely OpenVPN context, this change is correct.

angristan:
From what I understand, `ncp-ciphers` is interpreted as `data-ciphers` in 2.5. So we can keep it for compatibility with 2.4, right?

TinCanTech:
If 2.4 support is important then I would advise something like:
```
case "$(openvpn version)" in
2.4) ncp-ciphers foo
2.5|2.6) Nothing
esac
```
The default for `--data-ciphers` is `AES-256-GCM:AES-128-GCM`, which I use in my servers by not specifying `--data-ciphers`.

There is a warning about openvpn 2.4 selecting `BF-CBC` by default but you already have that covered.

PR #1010, opened 2022-07-12
feat: add flag for creation or not of new client after install

Podesta:
Add a flag `NEW_CLIENT` so that the user can choose whether or not he wishes to create a new user after installation.
It is specially useful on headless installations, when upgrading to a different server, but keeping old credentials.
It does not change any defaults, so if no flag is passed, it still creates the new user.


angristan (approved):
Thanks, I removed the prompt because I think this is an edge case, most people probably want to generate a client when running the script, so I'm keeping it that way

PR #1022, opened 2022-08-11
OpenWrt compatibility

LeviPO:
Slight changes to some commands to make them compatible with OpenWrt
Added openwrt in ifs with the OS condition
Special way of restarting services and adding rules to the firewall (taken from OpenWrt docs) 
Insignificant changes such as comments and echo outputs


angristan:
Thanks for this! Unfortunately, I'd rather keep support for systemd-based Linux distributions for now!

PR #1032, opened 2022-09-01
Fixed easyRSA asking for a password during setup

hvanmegen:
EasyRSA has been updated, also, the way the vars were stored as incorrect. This patch fixes it.


hvanmegen:
hm, found a problem during additional testing.. nevermind this

TinCanTech:
If the problem is with EasyRSA then I may be able to help.

PR #1033, opened 2022-09-02
Solve can not add the same client after revokeClient

jackielee524:


angristan:
This should be solved now!

PR #1040, opened 2022-09-16
Add multiserver capability

aleixq:
Adds multiple openvpn server capability, a new variable $SERVER_ID can be passed  it will create then the conf in /etc/openvpn/server.$SERVER_ID.conf and all server files will be placed in /etc/openvpn/server.$SERVER_ID/ .

for example `SERVER_ID=my_openvpn ./openvpn-install.sh`

It covers the server creation, the removal and the client creation.

(note: also changes the way how it proposes the public ip in install server questions. it tries to detect not private ips and if cannot find it fallbacks to private ips)

There are some pending tasks: 

## The ipv6 address for tunnel. 

The ipv4 for the tunnel is defined taking the most great second octet found in all servers configs: for example if there's a conf with 10.10.0.1 it will create the next openvpn server using 10.11.0.1. 
This logic is not implemented with ipv6.

## unbound

unbound part, as I don't know if it can accept multiple servers.


angristan:
Hi! Thanks for this, it's a cool change, but I'm not willing to keep on supporting this right now. I hope you understand. 🙏

PR #1054, opened 2022-10-28
Remove deprecated option

RewardedIvan:
OpenVPN will soon remove the `--secret` option in genkey, and it will be replace with `secret`


angristan:
Thanks. Any source on that? What about backward compatibility?

angristan:
Done in https://github.com/angristan/openvpn-install/pull/1059

PR #1060, opened 2022-11-10
Add variables to set expiration of CA certificate and client certificates

nodev11:
Hello there,

I've simply added two variables to set expiration of CA certificate and client certificates. 
Maybe the small modification is useful for somebody out there.

Cheers Frogger


angristan:
It would better follow the other configuration options on this, and support it as an env variable and add it the installation questions https://github.com/angristan/openvpn-install/blob/38d3bf9afa0733245a6b2eb7e149aaed54cddc67/openvpn-install.sh#L613-L642

angristan:
👋 I implemented certificate renewal for both server and clients certificates in https://github.com/angristan/openvpn-install/pull/1328. Sorry it took so long!

PR #1066, opened 2022-11-30
Fixed issue with ifconfig.co

starek4:
Hi all,

I am using this script on some VMs in Azure. It worked quite well for a while but stopped working yesterday. The issue is that when obtaining the public IP address, the `https://ifconfig.co` is now behind the Cloudfare, so the CURL request to it fails with HTTP code 503 and in response is this HTML instead of an IP address:
image

This PR comes with a fallback, whenever the `https://ifconfig.co` will not be useable for any reason, it will utilize the Google DNS service for those purposes.

I have tested it on Azure VMs and locally and it is working just fine.


angristan:
Thanks for this!

How about replacing ifconfig entirely by the `dig` call?

starek4:
I would not replace it entirely, it can also fail in some circumstances, let's use it as a fallback.

angristan (approved):

Issue #989, opened 2022-03-30
Add support of creating a client with the name of a deleted client


hcye:
### Make sure your check these beforehand!

- [X] Issues - https://github.com/angristan/openvpn-install/issues
- [X] README and FAQ - https://github.com/angristan/openvpn-install
- [X] Wiki - https://github.com/angristan/openvpn-install/wiki
- [X] Discussions - https://github.com/angristan/openvpn-install/discussions

### Server OS

ubuntu 20.04

### What is the bug?

add The following three lines after openvpn-install.sh line 1185 to fix this bug

         line=`grep -n R /etc/openvpn/easy-rsa/pki/index.txt`
         line_num=`echo ${line}  | cut -d ':' -f 1`
         sed -i  "${line_num}d" /etc/openvpn/easy-rsa/pki/index.txt

### Relevant log output

```shell
no output
```

angristan:
PR: https://github.com/angristan/openvpn-install/pull/1080

angristan:
This is supported now!

PR #1080, opened 2023-01-03
clear user to allow create a new one with the same name

MauricioFesta:


TinCanTech:
Deleting records from `index.txt` breaks CRL generation.

TinCanTech:
In order to recreate a client of the same name, `index.txt.attr` must be updated.

I have created an issue for this:
https://github.com/OpenVPN/easy-rsa/issues/853

angristan:
Thanks @TinCanTech! Is there a way to handle https://github.com/angristan/openvpn-install/issues/989 until https://github.com/OpenVPN/easy-rsa/issues/853 is tackled?

TinCanTech:
Add `unique_subject = no` to `index.txt.attr`.

Or run command `easyrsa upgrade ca`.
This can be used at any time after `init-pki`.

PR #1083, opened 2023-01-12
Update FAQ with IPv6 prefix policies for ULA addresses

gits7r:
Most operating systems will prefer IPv4 rather than IPv6 ULA, which defeats the purpose of redirect-gateway ipv6. It's nothing the script can do to automate this.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -152,7 +152,6 @@ route 10.0.0.0 255.0.0.0
 
 So for example - here it would route all traffic of `10.0.0.0/8` to the vpn. And the rest through the internet.
 
-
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -151,3 +151,26 @@ route 10.0.0.0 255.0.0.0
 ```
 
 So for example - here it would route all traffic of `10.0.0.0/8` to the vpn. And the rest through the internet.
+
+
```
angristan:
```suggestion
```

angristan (approved):
Thanks for contributing!

PR #1089, opened 2023-01-17
Update openvpn-install.sh

nu11secur1ty:
This is a significant UPDATE =) If you decide you can deploy this, dear friend, please, I will be glad you decide to do it. It is better to store the user configs in the inode of the OpenVPN ;) 
KR


angristan:
I would rather not change this, it's been working great so far

PR #1132, opened 2023-03-17
implement user deletion functionality

c0d3sling3r:
I wrote a function in the main script to remove a client information from `/etc/openvpn` directory and add an option in the `manageMenu` to operate the function.

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -1156,6 +1156,41 @@ function newClient() {
 	exit 0
 }
 
+function removeClient() {
+	CLIENTCOUNT=$(tail -n +2 /etc/openvpn/easy-rsa/pki/index.txt | wc -l)
+
+	echo "Select the existing client to delete:"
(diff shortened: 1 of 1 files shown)


TinCanTech:
This will cause revoked certificates to be allowed to connect to the VPN.

Please, do not edit EasyRSA `index.txt`, unless you know what you are doing.

c0d3sling3r:
It seems I'm missing something. How could it be possible to connect while all related files are removed and the `index.txt` is updated?

TinCanTech:
If you edit `index.txt` then you corrupt your CA database.

As for your other question, please learn how `X509` certificates work.

c0d3sling3r:
Thanks. I'll read about that.

Do you have any suggestions to remove a client in a clean way? I mean, is there any convenient way to do so? Probably you have thought about it before and it just cannot be done easily.

angristan:
The revoke function should be enough

Issue #1127, opened 2023-02-10
After rebooting the server, OpenVPN is not working.


realgagangupta:
Hello Team,

I have configured OpenVPN on my Ubuntu server, and everything is going fine, but when my server restarts, our vpn clients are not able to access the internet, although the VPN is still connected.

mindcont:
same

CHIP0K:
check it out iptables masquerading rules and IP forwarding

```bash
root@tf-ovpnxortcp-radius-blr1-5:/home/chip$ iptables-save | grep MASQUERADE
-A POSTROUTING -s 192.168.255.0/24 -o eth0 -j MASQUERADE
-A POSTROUTING -s 192.168.254.0/24 -o eth0 -j MASQUERADE
```

and
```bash
root@tf-ovpnxortcp-radius-blr1-5:/home/chip$ sysctl net.ipv4.ip_forward
net.ipv4.ip_forward = 1
```

so, my sysctl config


```bash
cat > /etc/sysctl.d/ovpnxor.conf <<EOF
net.ipv4.ip_forward = 1
net.ipv4.ip_no_pmtu_disc = 1
net.ipv4.conf.all.rp_filter = 1
net.ipv4.conf.all.accept_redirects = 0
net.ipv4.conf.all.send_redirects = 0
net.ipv6.conf.all.disable_ipv6 = 1
net.ipv6.conf.default.disable_ipv6 = 1
net.ipv6.conf.lo.disable_ipv6 = 1
net.ipv6.conf.SERVER_PUB_NIC.disable_ipv6 = 1
EOF
```

wid-get:
Hi, @gagangupta1404 @mindcont . I had the same issue. Edit /etc/systemd/system/iptables-openvpn.service and change line `Before=network-online.target` to `After=network-online.target`. Save the file, reboot and check.

gits7r:
@wid-get that's odd - on Debian Bullseye it works with `Before=network-online.target` , after every reboot all iptables and ip6tables rules are loaded fine. Wonder if it's something platform dependent?

angristan:
Fixed by https://github.com/angristan/openvpn-install/pull/1140

PR #1140, opened 2023-03-29
fix: use After=network-online.target for iptables service

wid-get:
fixes not executing add-openvpn-rules.sh after OS reboot.
systemctl shows service as _running_ but no iptables rules added to list. 
Also this fixes issue https://github.com/angristan/openvpn-install/issues/1127


gits7r:
Why does it usually work **without** this change and apparently just in some scenarios iptables / ip6tables rules are not loaded on reboot and this one line change is needed? I just want to make sure we don't merge something that will break the current functionality, since the script is working/loading on reboot without this PR in many environments. Thank you.

angristan (approved):

Issue #1158, opened 2023-06-22
Don't get public ip again if we already have it


afarah1:
**Issue**

With `AUTO_INSTALL=y` the public IP is obtained [here](https://github.com/angristan/openvpn-install/blob/master/openvpn-install.sh#L629). The installer then proceeds to get an IP again [here](https://github.com/angristan/openvpn-install/blob/master/openvpn-install.sh#L233) and if it's behind NAT, attempts to obtain the public IP again, this time using a different method, [here](https://github.com/angristan/openvpn-install/blob/master/openvpn-install.sh#L247). This second attempt is unnecessary if the first already succeeded. Furthermore, it is timing out as of now (tested on an EC2 and my local network), failing the script.

**Proposed change**

Do not attempt to get the public IP a second time if it was already successfully obtained the first time. I'll open a PR to that effect.

PR #1159, opened 2023-06-22
dont get ip if we already have it

afarah1:
Proposed change for https://github.com/angristan/openvpn-install/issues/1158


afarah1:
I've pushed a new commit switching the preferred methods to get the public ip as the previously second one seems to be more reliable. This way we also avoid slow retries unless we absolutely need them.

angristan:
Thanks for the PR, all of this is available now :)

PR #1174, opened 2023-10-05
Support aliyun like centos

mghayour:
Hi there,
I was unable to install this script on Aliyun (the custom linux for alibaba cloud service)
Aliyun is like centos and current version is 3
So i added this to os detection to work like centos and it worked for me.


angristan:
Sorry, I can't test it, so I won't support it: https://github.com/angristan/openvpn-install/pull/1322#discussion_r2602803474

Issue #974, opened 2022-01-31
Handle certificate expiration


olivamauricio:
Dear, I installed the script and I have the whole environment working, but I don't know when the certificates expire. Through the command below I verified that the ca.crt certificate has a period of 10 years to expire. But the server certificate is only 1 year old and will expire in the next few months.
How do I use the script to delay server certificate validity?

# openssl x509 -in **ca.crt** -dates -noout
notBefore=Jul 26 16:59:50 2019 GMT
notAfter=Jul 23 16:59:50 2029 GMT

# openssl x509 -in **server_p---------------.crt** -dates -noout
notBefore=Jul 26 16:59:50 2019 GMT
notAfter=Jul 10 16:59:50 2022 GMT

I imagine the server will stop working on Jul 10/2022. How to solve this problem? Thanks.

computeus:
I am also looking for an answer to this issue as my server certificate has expired this morning. Should I reinstall the OpenVPN server or is there a less disruptive solution?

katmai:
same thing here, the cert expired and nobody can connect. do we have to reinstall and reissue all clients profiles?

nxtreaming:
+1
same here

BigNerd95:
To fix the script you have to prepend `EASYRSA_CERT_EXPIRE=3650`  to:

https://github.com/angristan/openvpn-install/blob/master/openvpn-install.sh#L739

https://github.com/angristan/openvpn-install/blob/master/openvpn-install.sh#L1086

https://github.com/angristan/openvpn-install/blob/master/openvpn-install.sh#L1090

So for example it will become:
`EASYRSA_CERT_EXPIRE=3650 ./easyrsa build-server-full "$SERVER_NAME" nopass`

BigNerd95:
To fix the certificate of the server:
```bash
cd /etc/openvpn/easy-rsa/
rm pki/reqs/server_X194SFMGqcUxbZkB.req
rm pki/private/server_X194SFMGqcUxbZkB.key
rm pki/issued/server_X194SFMGqcUxbZkB.crt
EASYRSA_CERT_EXPIRE=3650 ./easyrsa build-server-full server_X194SFMGqcUxbZkB nopass
EASYRSA_CRL_DAYS=3650 ./easyrsa gen-crl
cp pki/crl.pem /etc/openvpn
cp pki/issued/server_X194SFMGqcUxbZkB.crt /etc/openvpn
[...]

PR #1235, opened 2024-07-12
Set client and server certificates validity to 10 years

angristan:
Prevent https://github.com/angristan/openvpn-install/issues/974


deafbabe:
Thank you for this, I've been doing it on my servers manually for years. One question tho, why did you chose to insert variable into command line instead of adding on line 720/723 `echo "set_var EASYRSA_CA_EXPIRE=3650" >>vars ` ? That is the approach that I've been taking and it seems much easier, but if it is bad I'd appreciate to know.

Issue #1241, opened 2024-08-07
Script does not work when seeip.org is unreachable


gbnk0:
```
./openvpn-install.sh

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0^C
```

It does a curl to ip.seeip.org . When seeip.org is down or unreachable it dos not work. I think it will tell something like "unable to get the public ip" or anything else.

I thin the concerned lines are:

```
			if ! PUBLIC_IP=$(curl -f --retry 5 --retry-connrefused https://ip.seeip.org); then
			if ! PUBLIC_IP=$(curl -f --retry 5 --retry-connrefused -4 https://ip.seeip.org); then
```
Regards

dolbykoray:
Senin sorunu`apt install iftop` çözebilir denedin mi? Belkide sorununu yanlış anladım selamlar.

gbnk0:
The curl is trying indefinitely to hit ip.seeip.org, so in this case you can't access to the options for creating new vpn clients.
Update:

This happens only when i pass env vars in my shell before:
```
export MENU_OPTION=1 PASS=1 CLIENT=myclient AUTO_INSTALL=y ENDPOINT=servrer.publicip.org IPV6_SUPPORT=n PORT_CHOICE=2 PORT=4567 PROTOCOL_CHOICE=2
```

Regards,

dolbykoray:
Can you solve my problem? or give me some ideas?   [https://github.com/angristan/openvpn-install/issues/1242](url)

ralphg6:
I did a PR fixing it #1243.

Please @angristan, could you put your eyes on it? 

Now the script isn't working because 'ip.seeip.org' is unreachable now.

fmuscetra:
I have the same issue. My Terraform template that uses the Angristan shell script does not work anymore and it hangs when the script trying to connect to ip.seeip.org. 
It worked for months without any problem. 
The problem appears when AUTO_INSTALL = y

ralphg6:
The PR that I mentioned fix it!! Let's await some maintainer to put your eyes on it!


@fmuscetra, if you REALLY need it now, and you can't await the PR reviewing, you can use my fork temporarily! 

https://github.com/ralphg6/openvpn-install

xiahare:
I submit a PR for this problem.
[...]

PR #1243, opened 2024-08-10
Fix Public IP detection - Fix issue when seeip.org is unreachable #1241

ralphg6:
The script does work when seeip.org is unreachable, so I changed the policy to define the public IP.

It solves the issue #1241

### Added

- Timeout limit on each try to solve the IP to avoid long waits;
- Extra public IP providers as failovers;

### Changed

- the script only will try to solve an IP if the ENDPOINT is empty;

### Removed

- Unneeded output of CURL commands on IP detection;
- Duplicated code;

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -216,6 +216,45 @@ access-control: fd42:42:42:42::/112 allow' >>/etc/unbound/openvpn.conf
 	systemctl restart unbound
 }
 
+function resolvePublicIP() {
+	# IP version flags, we'll use as default the IPv4
+	CURL_IP_VERSION_FLAG="-4"
+	DIG_IP_VERSION_FLAG="-4"
+
+	# Behind NAT, we'll default to the publicly reachable IPv4/IPv6.
+	if [[ $IPV6_SUPPORT == "y" ]]; then
+		CURL_IP_VERSION_FLAG=""
+		DIG_IP_VERSION_FLAG="-6"
+	fi
+
+	# If there is no public ip yet, we'll try to solve it using: https://ip.seeip.org
+	if [[ -z $PUBLIC_IP ]]; then
+		PUBLIC_IP=$(curl -f -m 5 -sS --retry 5 --retry-connrefused $CURL_IP_VERSION_FLAG https://ip.seeip.org 2>/dev/null)
+	fi
+
+	# If there is no public ip yet, we'll try to solve it using: https://ifconfig.me
+	if [[ -z $PUBLIC_IP ]]; then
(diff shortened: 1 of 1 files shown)


ralphg6:
Solves #1241

alaz (approved):

openvpn-install.sh
```diff
@@ -216,6 +216,45 @@ access-control: fd42:42:42:42::/112 allow' >>/etc/unbound/openvpn.conf
...
+
+	# Behind NAT, we'll default to the publicly reachable IPv4/IPv6.
+	if [[ $IPV6_SUPPORT == "y" ]]; then
+		CURL_IP_VERSION_FLAG=""
+		DIG_IP_VERSION_FLAG="-6"
+	fi
+
+	# If there is no public ip yet, we'll try to solve it using: https://ip.seeip.org
+	if [[ -z $PUBLIC_IP ]]; then
+		PUBLIC_IP=$(curl -f -m 5 -sS --retry 2 --retry-connrefused $CURL_IP_VERSION_FLAG https://ip.seeip.org 2>/dev/null)
```
angristan: cf https://github.com/angristan/openvpn-install/pull/1252

```suggestion
	# If there is no public ip yet, we'll try to solve it using: https://api.seeip.org
	if [[ -z $PUBLIC_IP ]]; then
		PUBLIC_IP=$(curl -f -m 5 -sS --retry 2 --retry-connrefused $CURL_IP_VERSION_FLAG https://api.seeip.org 2>/dev/null)
```

angristan (changes requested):

angristan:
Thanks 🙏 I updated the seeip.org endpoint

openvpn-install.sh
```diff
@@ -216,6 +216,45 @@ access-control: fd42:42:42:42::/112 allow' >>/etc/unbound/openvpn.conf
...
+		PUBLIC_IP=$(curl -f -m 5 -sS --retry 2 --retry-connrefused $CURL_IP_VERSION_FLAG https://api.ipify.org 2>/dev/null)
+	fi
+
+	# If there is no public ip yet, we'll try to solve it using: ns1.google.com
+	if [[ -z $PUBLIC_IP ]]; then
+		PUBLIC_IP=$(dig $DIG_IP_VERSION_FLAG TXT +short o-o.myaddr.l.google.com @ns1.google.com | tr -d '"')
+	fi
+
+	if [[ -z $PUBLIC_IP ]]; then
+		>&2 echo "Couldn't solve the public IP"
```
angristan:
```suggestion
		echo >&2 echo "Couldn't solve the public IP"
```

angristan:
Thanks again, this is great 😄

PR #1250, opened 2024-10-12
Allow custom certificate duration when creating clients

omidshojaee:
For those who need it, the script asks for how many days the new client should be valid. This defaults to 3650 days. Then it sets the ```EASYRSA_CERT_EXPIRE``` variable accordingly.

This script is meant to be simple which means it is for those who are not tech-savvy to handle the complex task of installing and configuring OpenVPN.

However if the user has a large number of clients and all of them are valid for 10 years, it is very hard to keep track of them. This PR helps them to set a reasonable validity period, while the default is the same.


IllusionVK:
Hey nice enhancement @omidshojaee !

angristan (approved):
Thanks!

PR #1252, opened 2024-10-18
url ip.seeip.org has been changed to api.seeip.org

xiahare:
the url of ip.seeip.org is not valid. It has been changed to api.seeip.org

Refer to https://seeip.org/


angristan (approved):

angristan:
Thanks!

PR #1273, opened 2024-12-23
Disable of creating first client during install openvpn, auto select …

skyatua:
…client when revoke flow


angristan:
Supported with https://github.com/angristan/openvpn-install/pull/1010

PR #1278, opened 2025-01-23
Add option to allow multiple devices per client profile (duplicate-cn)

Shahzainali:
Added duplicate-cn for connecting multiple clients using same .ovpn


dostmuhammaddoshi (changes requested):
It will help a lot

angristan (approved):
Thanks!

PR #1287, opened 2025-03-24
feat(ipv6): Add full IPv6-only mode with manual hostname/IP fallback …

Knogle:
…if no IPv4

- Detects absence of global IPv4 and prompts user for manual public IPv4 or hostname.
- If user declines, falls back to checking for public IPv6 address.
- If global IPv6 is available, offers IPv6-only mode as a valid setup path.
- Ensures $IP and $ENDPOINT are set appropriately in all scenarios.
- Updates public IP fallback logic to also handle IPv6-only configurations.
- Adds support for manual entry of public IPv6 hostname/IP when behind NAT or no IPv4 is available.

This improves support for IPv6-only server deployments and ensures usability in NAT64, DS-Lite, or cloud environments without native IPv4.

```
root@OVPN-Test:~# ./openvpn-install.sh 
Welcome to the OpenVPN installer!
The git repository is available at: https://github.com/angristan/openvpn-install

I need to ask you a few questions before starting the setup.
You can leave the default options and just press enter if you are OK with them.

IP address for OpenVPN to listen on: 2a00:XXXX:XXXX:XXXX:XXX:XXXX:XXXX:XXXX

Checking for IPv6 connectivity...

Your host appears to have IPv6 connectivity.

Do you want to enable IPv6 support (NAT)? [y/n]: y

What port do you want OpenVPN to listen to?
   1) Default: 1194
   2) Custom
   3) Random [49152-65535]
Port choice [1-3]: 1
```

### DIFF
--- a/openvpn-install.sh
+++ b/openvpn-install.sh
@@ -25,7 +25,7 @@ function checkOS() {
 			if [[ $VERSION_ID -lt 9 ]]; then
 				echo "⚠️ Your version of Debian is not supported."
 				echo ""
-				echo "However, if you're using Debian >= 9 or unstable/testing then you can continue, at your own risk."
+				echo "However, if you're using Debian >= 9 or unstable/testing, you can continue at your own risk."
 				echo ""
 				until [[ $CONTINUE =~ (y|n) ]]; do
 					read -rp "Continue? [y/n]: " -e CONTINUE
@@ -40,7 +40,7 @@ function checkOS() {
 			if [[ $MAJOR_UBUNTU_VERSION -lt 16 ]]; then
 				echo "⚠️ Your version of Ubuntu is not supported."
 				echo ""
-				echo "However, if you're using Ubuntu >= 16.04 or beta, then you can continue, at your own risk."
+				echo "However, if you're using Ubuntu >= 16.04 or beta, you can continue at your own risk."
 				echo ""
 				until [[ $CONTINUE =~ (y|n) ]]; do
 					read -rp "Continue? [y/n]: " -e CONTINUE
@@ -60,7 +60,7 @@ function checkOS() {
 			if [[ ${VERSION_ID%.*} -lt 7 ]]; then
 				echo "⚠️ Your version of CentOS is not supported."
 				echo ""
-				echo "The script only support CentOS 7 and CentOS 8."
+				echo "The script only supports CentOS 7 and CentOS 8."
 				echo ""
 				exit 1
 			fi
@@ -70,7 +70,7 @@ function checkOS() {
 			if [[ ! $VERSION_ID =~ (8) ]]; then
 				echo "Your version of Oracle Linux is not supported."
 				echo ""
-				echo "The script only support Oracle Linux 8."
+				echo "The script only supports Oracle Linux 8."
 				exit 1
 			fi
 		fi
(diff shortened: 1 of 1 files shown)


m-primo:
Hello,
I have tested this change, and unfortunately it doesn't work on IPv6-only systems,
I'm unable to test further because GitHub itself doesn't support IPv6, so the installation halts on this step:

```
--2025-05-02 10:28:40--  https://github.com/OpenVPN/easy-rsa/releases/download/v3.1.2/EasyRSA-3.1.2.tgz
Resolving github.com (github.com)... 140.82.121.4
Connecting to github.com (github.com)|140.82.121.4|:443... failed: Network is unreachable.
```

Tried with ping, curl and other stuff, it seems like github domain doesn't have any kind of IPv6 support (that's super weird for a big company like github!). (i have searched about this and seems like they are working on it since 2022 lol)

and then follows dozens of errors:

```
gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
./openvpn.sh: line 801: ./easyrsa: No such file or directory
./openvpn.sh: line 802: ./easyrsa: No such file or directory
./openvpn.sh: line 808: ./easyrsa: No such file or directory
./openvpn.sh: line 809: ./easyrsa: No such file or directory
2025-05-02 10:28:40 DEPRECATED OPTION: The option --secret is deprecated.
2025-05-02 10:28:40 WARNING: Using --genkey --secret filename is DEPRECATED.  Use --genkey secret filename instead.
cp: cannot stat 'pki/ca.crt': No such file or directory
cp: cannot stat 'pki/private/ca.key': No such file or directory
cp: cannot stat 'pki/issued/server_1xCfbb8s10mBvTTW.crt': No such file or directory
cp: cannot stat 'pki/private/server_1xCfbb8s10mBvTTW.key': No such file or directory
cp: cannot stat '/etc/openvpn/easy-rsa/pki/crl.pem': No such file or directory
chmod: cannot access '/etc/openvpn/crl.pem': No such file or directory
```

then it ends with:

```
Created symlink /etc/systemd/system/multi-user.target.wants/openvpn@server.service → /etc/systemd/system/openvpn@.service.
Job for openvpn@server.service failed because the control process exited with error code.
See "systemctl status openvpn@server.service" and "journalctl -xeu openvpn@server.service" for details.
Created symlink /etc/systemd/system/multi-user.target.wants/iptables-openvpn.service → /etc/systemd/system/iptables-openvpn.service.
```

This is on an IPv6-only machine, no IPv4 available at all.

I understand this is Github's fault to not support IPv6, but in order for this change to be truly IPv6-only, we need to find another way of downloading EasyRSA over IPv6 (or dual-stack), probably a mirror. I think we could fix it manually by downloading it somehow and then restart the openvpn service, but this would still make this script not ipv6-only support.

I'm happy to continue testing this until it works on IPv6-only systems.

Knogle:
Oh hey, i hope you are doing fine :)
It wasn't really meant for systems which are only IPv6 capable, but rather systems that are only publicly reachable through IPv6 like hosts beyond CGNAT. So they don't have a public IPv4 exposed to be reachable on.

m-primo:
It's okay, no problem, I'll try to look for solutions anyway.

PR #1374, opened 2025-12-13
fix: use After=network-online.target for iptables service

angristan:
- Change iptables-openvpn systemd service to start after network-online.target instead of before
- Ensures the network interface exists when NAT rules are applied


angristan:
Dupe of https://github.com/angristan/openvpn-install/pull/1140

PR #1450, opened 2026-01-07
Add AppArmor compatability

davidrozen76:
* Checks whether AppArmor is installed and enabled.
* Adds the proper AppArmor profile for OpenVPN.


angristan:
Thanks! I added support in https://github.com/angristan/openvpn-install/pull/1469

PR #1462, opened 2026-01-29
Added support for Linux Mint to the installOpenVPNRepo() function

davycrocket144:
Added support for Linux Mint. 
Tested and Verified code on my Linux Mint Laptop and MiniPC as well as my Headless Debian 13 Server. 
Passed all Workflows.


davycrocket144:
Seems there is an issue with the Ubuntu 24.04 fingerprint test, I ran it on my own fork and it failed at first, but when I re-ran the test it passed.

PR #1515, opened 2026-09-19
Add custom output directory for client revoke

yurymuski:
Issue: `client add`  already supports storing .ovpn files in a custom directory with --output, but `client revoke` didn't remove those files.

Added `--output-dir` to `client revoke` so it can also remove .ovpn files stored in a custom directory.

For example:

`./openvpn-install.sh client revoke alice --force --output-dir /srv/openvpn/clients`

This will revoke the client and remove:

`/srv/openvpn/clients/alice.ovpn`

Also updated the README and command help with the new option.

Existing revoke behavior stays unchanged when `--output-dir` isn't provided.


yurymuski:
failed test are unrelated to the PR, and seems I can not restart them

